PR #187, opened 2022-06-15
DXCDT-155: Fix subsequent updates to partial `refresh_token` object

willvedd:
## Description

As noted in #24, there is an issue when applying client configurations when not all properties of a refresh token are set. The initial config is applied appropriately, but subsequent applies are met with this general error:

```
 refresh_token {
          - idle_token_lifetime          = 1296000 -> null
          - infinite_idle_token_lifetime = true -> null
          - infinite_token_lifetime      = true -> null
          - token_lifetime               = 2592000 -> null
            # (3 unchanged attributes hidden)
        }
        # (1 unchanged block hidden)
        
Error: 400 Bad Request: Neither infinite_token_lifetime nor infinite_idle_token_lifetime can be true when token expiration type is 'non-expiring'
```

What's occurring is that several of these properties within a client's `refresh_token` object are set to defaults if not explicitly declared. By setting `Computed: true` to each of these properties, we signal to the provider that the default values should be used instead of remaining empty.

## Checklist

#### Auth0 Code of Conduct

- [x] Yes
- [ ] Not needed

- [x] Not needed


sergiughf (approved):
Great work fixing this 💪🏻

Widcket (approved):

PR #193, opened 2022-06-16
Setting session liftetime values as non-computed

willvedd:
## Description

As noted in #150 , the session lifetime values, `session_lifetime` and `idle_session_lifetime` would default to zero when not returned by the API. These values cannot be zero and if state was migrated to another tenant, the TF plan would fail. In most cases, these session lifetime values are set, but may not be certain situations like newly-created tenants. By reverting these schema properties from `Computed: true` to `Computed: false`, we indicate that no default should be set if either value is not returned by the Management API.

## Checklist

#### Auth0 Code of Conduct

- [ ] Yes
- [ ] Not needed

- [ ] Not needed


willvedd:
Also worth mentioning that many of the tenant's values are erroneously set to `Computed: true`. It was done in [this PR](https://github.com/alexkappa/terraform-provider-auth0/pull/194/files) which made many schema properties set as computed when they likely shouldn't be. This issue was caused by that massive change and we should fix them all eventually. However, I decided to just focus on this single issue and we can address the rest at a later point in time.

Widcket:
CI is failing.

sergiughf (approved):

PR #240, opened 2022-07-14
Connection flatten functions to return diagnostics slice instead of error

willvedd:
## Description

Small refactor of the connection resource's flatten functions to return `diag.Diagnostics` instead of a regular error. What this does is allow more specific diagnostic logging to exist alongside the domain-specific code. Otherwise, we would need to maintain some type of mapping between a custom error type and the diagnostic logs within the higher-level flatten function.

## Checklist

#### Auth0 Code of Conduct

- [ ] Yes
- [ ] Not needed

- [ ] Not needed


auth0/structure_auth0_connection_test.go
```diff
@@ -0,0 +1,15 @@
+package auth0
+
+import "testing"
+
+func TestFlattenConnectionOptions(t *testing.T) {
```
willvedd, line 5: I always wondered why we didn't unit test the flatten function because there is quite a bit of logic and complexity wrapped up in them. However, this test is just asserting the simple case of a `nil` options struct which should return `nil`.

sergiughf (dismissed):

sergiughf (approved):

PR #562, opened 2023-04-25
Don't override imported connection_client id

WesleyKlop:
### 🔧 Changes

This PR changes the `importConnectionClient` function which handles the importing of `auth0_connection_client` resources. When repeatedly importing this resource (like is possible when using a tool like Pulumi) the imported connection client has a new id every time even though it has not changed. This causes the IaC tool to want to recreate the resource, which breaks your environment. Using just the combination of `connection_id:client_id` is unique enough and should be fine to use the identifier for this resource, since I don't think there is a situation where duplicates would exist. (and if they do, that sounds like a misconfiguration)

The only change is that the internal id is now the same each time the same resource is imported.
 

### 🔬 Testing

This can be tested by importing connection clients!

### 📝 Checklist

- [ ] ~I have added documentation for all new/changed functionality (or N/A)~

### DIFF
--- a/internal/auth0/attackprotection/data_source.go
+++ b/internal/auth0/attackprotection/data_source.go
@@ -24,6 +24,7 @@ func dataSourceSchema() map[string]*schema.Schema {
 }
 
 func readAttackProtectionForDataSource(ctx context.Context, data *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	// This resource is not identified by an id in the Auth0 management API.
 	data.SetId(id.UniqueId())
 	return readAttackProtection(ctx, data, meta)
 }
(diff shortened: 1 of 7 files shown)


sergiught (commented):
Appreciate the contribution @WesleyKlop! We weren't aware of pulumi internal mechanics, however we have other such similar import mechanics across other resources like `auth0_organization_connection` and `auth0_organization_member`. If we're going this path would you mind altering the import for those resources as well? So we have consistency? 🙏🏻

WesleyKlop:
Yes ofcourse, will do!

WesleyKlop:
> Appreciate the contribution @WesleyKlop! We weren't aware of pulumi internal mechanics, however we have other such similar import mechanics across other resources like `auth0_organization_connection` and `auth0_organization_member`. If we're going this path would you mind altering the import for those resources as well? So we have consistency? 🙏🏻

@sergiught I changed it! :)

sergiught (approved):
Thanks a lot for the contribution @WesleyKlop 💯 Great work and appreciate going the extra mile on this one 💪🏻

WesleyKlop:
Thank you very much! :D

PR #574, opened 2023-05-10
DXCDT-427: User-permission resource (2/2)

willvedd:
### 🔧 Changes

This PR introduces a `auth0_user_permission` resource that enables the management of user permissions (detached from role) with a 1-1 relationship. 

This work depends on #572 which is used for testing to assert that this resource functions as expected.

### 📚 References

Dependent PR: #572 

### 🔬 Testing

Added several integration test cases.

### 📝 Checklist


internal/auth0/user/resource_permission.go
```diff
@@ -0,0 +1,169 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Name of the permission.",
+			},
+			"resource_server_identifier": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Identifier (ID) of the resource server that the permission is associated with.",
```
sergiught, line 37: The identifier is not the resource server ID, they're 2 separate things. However, is the ID also accepted by the API?

image

internal/auth0/user/resource_permission.go
```diff
@@ -0,0 +1,169 @@
...
+
+// NewPermissionResource will return a new auth0_connection_client resource.
+func NewPermissionResource() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"user_id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "ID of user to associate the permission to.",
```
sergiught, line 25:
```suggestion
				Description: "ID of the user to associate the permission to.",
```

internal/auth0/user/resource_permission.go
```diff
@@ -0,0 +1,169 @@
...
+	_ context.Context,
+	data *schema.ResourceData,
+	_ interface{},
+) ([]*schema.ResourceData, error) {
+	rawID := data.Id()
+	if rawID == "" {
+		return nil, fmt.Errorf("ID cannot be empty")
+	}
+
+	if !strings.Contains(rawID, "::") {
```
sergiught, line 153: In all the other resource pairs we use ":" as the separator, what made us use "::" over here? 🤔 Is it some kind of limitation when used with the user ID?

internal/auth0/user/resource_permission.go
```diff
@@ -0,0 +1,169 @@
...
+			return nil
+		}
+		return diag.FromErr(err)
+	}
+
+	data.SetId("")
+	return nil
+}
+
+func importUserPermission(
```
sergiught, line 143: Happy to go forward with this, but it would be nice if we have a follow up PR where we change the schema.ImportResourceIDPair func to accept a variadic list of resource attribute names.

internal/auth0/user/resource_permission.go
```diff
@@ -0,0 +1,169 @@
...
+			StateContext: importUserPermission,
+		},
+		Description: "With this resource, you can manage user permissions.",
+	}
+}
+
+func createUserPermission(ctx context.Context, data *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	api := meta.(*management.Management)
+
+	userId := data.Get("user_id").(string)
```
sergiught, line 63:
```suggestion
	userID := data.Get("user_id").(string)
```

ID, URL, HTTP etc. are special names in golang and should be all capitalized.

internal/auth0/user/resource_permission.go
```diff
@@ -0,0 +1,169 @@
...
+	if err := api.User.AssignPermissions(userId, []*management.Permission{
+		{
+			ResourceServerIdentifier: &resourceServerId,
+			Name:                     &permissionName,
+		},
+	}); err != nil {
+		return diag.FromErr(err)
+	}
+
+	data.SetId(userId + resourceServerId + permissionName)
```
sergiught, line 79:
```suggestion
	data.SetId(userId + ":"+ resourceServerId + ":" + permissionName)
```

internal/auth0/user/resource_permission.go
```diff
@@ -0,0 +1,169 @@
...
+		if mErr, ok := err.(management.Error); ok && mErr.Status() == http.StatusNotFound {
+			data.SetId("")
+			return nil
+		}
+		return diag.FromErr(err)
+	}
+
+	for _, p := range existingPermissions.Permissions {
+		if p.GetName() == permissionName && p.GetResourceServerIdentifier() == resourceServerId {
+			data.Set("description", p.GetDescription())
```
sergiught, line 102: `data.Set` returns an error, so we should check for it with the multierror here

internal/auth0/user/resource_permission_test.go
```diff
@@ -0,0 +1,128 @@
...
+	resource_server_identifier = auth0_resource_server.resource_server.identifier
+	permission = "create:foo"
+}
+`
+
+const givenAUserWithNoPermissions = testAccUserEmpty
+const givenAUserWithOnePermission = givenAUserWithNoPermissions + givenAUserPermission
+const givenAUserWithTwoPermissions = givenAUserWithOnePermission + givenAnotherUserPermission
+
+const testAccUserPermissionNoneAssigned = givenAResourceServerWithPermissions + givenAUserWithNoPermissions + `data "auth0_user" "user" {
```
sergiught, line 51: We don't need to use a data source to retrieve the user details, we already have access to the user resource.

internal/auth0/user/resource_permission.go
willvedd: Can we enforce with linting?

sergiught: Absolutely, feel free to leave as is in this PR if you want and we can have another follow up one with the updates to the golangci-lint.

internal/auth0/user/resource_permission.go
sergiught: Ouch 😢 , we have all the other resource pair getting imported just through ":", it would be great if we could make all of them consistent. Perhaps introduce another separator? And gate it under the v1 if we need a breaking change? Or make this one a bit smarter when it parses permission names?

internal/auth0/user/resource_permission.go
willvedd: Fixed.

internal/auth0/user/resource_permission.go
willvedd: I'm ok to use `::` for all or only special cases but would like to avoid introducing this breaking change into v0.

examples/resources/auth0_user_permission/resource.tf
```diff
@@ -0,0 +1,33 @@
...
+  nickname        = "some.nickname"
+  email           = "test@test.com"
+  email_verified  = true
+  password        = "passpass$12$12"
+  picture         = "https://www.example.com/a-valid-picture-url.jpg"
+}
+
+resource "auth0_user_permission" "user_permission_read" {
+  depends_on = [auth0_resource_server.resource_server, auth0_user.user]
+
```
sergiught, line 29:
```suggestion
```

Hint: It's unnecessary to add this here because both the user and resource server references are used below, so terraform will automatically imply that they're dependent upon.

examples/resources/auth0_user_permission/resource.tf
```diff
@@ -0,0 +1,33 @@
...
+  user_id                    = auth0_user.user.id
+  resource_server_identifier = auth0_resource_server.resource_server.identifier
+  permission                 = "create:foo"
```
sergiught, line 32:
```suggestion
  permission                 = tolist(auth0_resource_server.resource_server.scopes)[0]
```

Hint: You can also do this.

internal/auth0/user/resource_permission_test.go
```diff
@@ -0,0 +1,117 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: acctest.ParseTestName(testAccUserPermissionNoneAssigned, strings.ToLower(t.Name())),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("auth0_user.user", "permissions.#", "0"),
+				),
+			},
+			{
+				Config: acctest.ParseTestName(testAccUserPermissionOneAssigned, strings.ToLower(t.Name())),
+			},
```
sergiught, line 62:
```suggestion
			{
				Config: acctest.ParseTestName(testAccUserPermissionOneAssigned, strings.ToLower(t.Name())),
			},
```
This seems to be unnecessary as it is the exact same config as below.

internal/auth0/user/resource_permission_test.go
```diff
@@ -0,0 +1,117 @@
...
+resource "auth0_user_permission" "user_permission_create" {
+	depends_on = [ auth0_resource_server.resource_server, auth0_user.user ]
+
+	user_id = auth0_user.user.id
+	resource_server_identifier = auth0_resource_server.resource_server.identifier
+	permission = "create:foo"
+}
+`
+
+const testAccUserPermissionNoneAssigned = givenAResourceServerWithPermissions + testAccUserEmpty
```
sergiught, line 47: Because there's no `depends_on` within the `testAccUserEmpty` config, when running the `testAccUserPermissionNoneAssigned` scenario, sometimes the API call that runs first is the one to create the resource server and at other times it's the user, that makes our tests sometimes fail because the execution order of saved HTTP requests is not consistent. 

I would suggest having a 

```go
const givenAResourceServerWithPermissionsAndAUser = `
resource "auth0_resource_server" "resource_server" {
	name = "Acceptance Test - {{.testName}}"
	identifier = "https://uat.api.terraform-provider-auth0.com/{{.testName}}"
	scopes {
		value = "read:foo"
		description = "Can read Foo"
	}
	scopes {
		value = "create:foo"
		description = "Can create Foo"
	}
}

resource auth0_user user {
	depends_on = [ auth0_resource_server.resource_server ]

	connection_name = "Username-Password-Authentication"
	user_id = "{{.testName}}"
	username = "{{.testName}}"
	password = "passpass$12$12"
	email = "{{.testName}}@acceptance.test.com"
}
`
```

internal/auth0/user/resource_permission_test.go
```diff
@@ -0,0 +1,117 @@
...
+					resource.TestCheckResourceAttr("auth0_user_permission.user_permission_read", "permission", "read:foo"),
+					resource.TestCheckResourceAttr("auth0_user_permission.user_permission_read", "user_id", "auth0|testaccuserpermission"),
+					resource.TestCheckResourceAttr("auth0_user_permission.user_permission_read", "resource_server_identifier", "https://uat.api.terraform-provider-auth0.com/testaccuserpermission"),
+					resource.TestCheckResourceAttr("auth0_user_permission.user_permission_read", "resource_server_name", "Acceptance Test - testaccuserpermission"),
+					resource.TestCheckResourceAttr("auth0_user_permission.user_permission_read", "description", "Can read Foo"),
+				),
+			},
+			{
+				Config: acctest.ParseTestName(testAccUserPermissionTwoAssigned, strings.ToLower(t.Name())),
+			},
```
sergiught, line 81:
```suggestion

```

This seems to be unnecessary as it is the exact same config as below.

PR #582, opened 2023-05-16
DXCDT-431: `auth0_role_permission` resource

willvedd:
### 🔧 Changes

Introducing the `auth0_role_permission` resource which enables management of role permissions through a 1-1 relationship.

This PR shares similarities with the `auth0_user_permission` resource (#574) but with extra considerations for the already-existing `permissions` field on the role resource like the migration guide and deprecation.

### 📚 References

Similar PR: #574 

### 🔬 Testing

Added full suite of integration tests for this resource.

### 📝 Checklist


MIGRATION_GUIDE.md
```diff
@@ -66,14 +67,14 @@ resource "auth0_user" "user" {
...
 
-# Use the auth0_user_roles to manage a 1:many 
+# Use the auth0_user_roles to manage a 1:many
 # relationship between the user and its roles.
 resource auth0_user_roles user_roles {
   user_id = auth0_user.user.id
   roles = [auth0_role.admin.id]
 }
 
-# Or the auth0_user_roles to manage a 1:1
```
willvedd, line 76: Quick fix

internal/auth0/role/resource_permission_test.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	"strings"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+
+	"github.com/auth0/terraform-provider-auth0/internal/acctest"
+)
+
+const givenARole = `
```
sergiught, line 12:
```suggestion
const givenAResourceServerAndARole = `
```

internal/auth0/role/resource_permission_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+data "auth0_role" "role" {
+	role_id = auth0_role.role.id
+}
+`
+
+const testAccRolePermissionsNoneAssigned = givenARole
+
+const testAccRolePermissionsOneAssigned = givenARole + `
+resource "auth0_role_permission" "role_permission" {
```
sergiught, line 43: This now executes after the data source gets refreshed, which can lead to inconsistent data on the resource. The data source should get added after this resource executes and have it as a dependency. See: resource_role_test.go in the user pkg. Like that we don't need to have separate refresh state test steps.

internal/auth0/role/resource_permission_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+					resource.TestCheckResourceAttr("data.auth0_role.role", "permissions.0.name", "create:foo"),
+					resource.TestCheckResourceAttr("data.auth0_role.role", "permissions.0.resource_server_identifier", "https://uat.testaccrolepermission.terraform-provider-auth0.com/api"),
+				),
+			},
+			{
+				Config: acctest.ParseTestName(testAccRolePermissionsOneAssigned, strings.ToLower(t.Name())),
+			},
+			{
+				RefreshState: true,
```
sergiught, line 112:
```suggestion
				RefreshState: true,
```

image

Should this be here? I think we can move the checks in the config above if we use the data source as the last elements that gets refreshed.

willvedd: This needs to be here for tests to pass. The data source gets executed after (or at best, during) the deletions of the `auth0_role_permission` resources. AFAICT, there is no way to depend on a deleted resource.

sergiught: Ah, I see that now 👍🏻 , indeed the execution flow on a terraform apply is:

read and refresh state
creations
updates
deletions

so we need another refresh state after the deletion.

sergiught (approved):
💪🏻

PR #621, opened 2023-06-07
DXCDT-462: Adding optional `display_name` field to `auth0_trigger_action` resource

willvedd:
### 🔧 Changes

The initial iteration of the `auth0_trigger_action` resource (#612) did not include a display name field. This field is used to assign a distinguishable name to a specific instance of an action within a flow. This can be useful in cases where there are multiple instances of the same action with a flow, albeit configured differently. The `auth0_trigger_actions` resource does include this property so for API consistency, this PR adds the `display_name` property too.

### 📚 References

- First iteration of `auth0_trigger_action` #612 

### 🔬 Testing

Adding new test cases.

### 📝 Checklist


internal/auth0/action/resource_trigger_action.go
```diff
@@ -86,21 +94,20 @@ func createTriggerAction(_ context.Context, d *schema.ResourceData, m interface{
 		})
 	}
 
-	action, err := api.Action.Read(actionID)
-	if err != nil {
-		if mErr, ok := err.(management.Error); ok && mErr.Status() == http.StatusNotFound {
-			d.SetId("")
-			return nil
+	if d.GetRawConfig().GetAttr("display_name").IsNull() {
```
sergiught, line 97: Because the `display_name` is computed and is of type string it will never actually be `nil`, plus we're already having this value from line 70. This should be changed to:

```suggestion
	if displayName == "" {
```

internal/auth0/action/resource_trigger_action.go
```diff
@@ -86,21 +94,20 @@ func createTriggerAction(_ context.Context, d *schema.ResourceData, m interface{
...
+		action, err := api.Action.Read(actionID)
+		if err != nil {
+			return diag.FromErr(err)
```
sergiught, line 100: We should probably check for a 404 here.

internal/auth0/action/resource_trigger_action.go
```diff
@@ -111,6 +118,59 @@ func createTriggerAction(_ context.Context, d *schema.ResourceData, m interface{
...
+	trigger := d.Get("trigger").(string)
+	actionID := d.Get("action_id").(string)
+	displayName := d.Get("display_name").(string)
+
+	currentBindings, err := api.Action.Bindings(trigger)
+	if err != nil {
+		if mErr, ok := err.(management.Error); ok && mErr.Status() == http.StatusNotFound {
+			d.SetId("")
+			return nil
+		}
```
sergiught, line 133: We shouldn't ignore the 404 here.

internal/auth0/action/resource_trigger_action.go
```diff
@@ -111,6 +118,59 @@ func createTriggerAction(_ context.Context, d *schema.ResourceData, m interface{
...
+		return diag.FromErr(err)
+	}
+
+	found := false
+	var updatedBindings []*management.ActionBinding
+	for _, binding := range currentBindings.Bindings {
+		if binding.Action.GetID() == actionID {
+			d.SetId(trigger + "::" + actionID)
```
sergiught, line 141: We should never set the ID of a resource in an update func. 
```suggestion
```

internal/auth0/action/resource_trigger_action.go
```diff
@@ -111,6 +118,59 @@ func createTriggerAction(_ context.Context, d *schema.ResourceData, m interface{
...
+			found = true
+		} else {
+			updatedBindings = append(updatedBindings, &management.ActionBinding{
+				Ref: &management.ActionBindingReference{
+					Type:  auth0.String("action_id"),
+					Value: binding.Action.ID,
+				},
+				DisplayName: binding.DisplayName,
+			})
+		}
```
sergiught, line 158: If you continue in the if above, we don't need an else here so we can reduce nesting.

internal/auth0/action/resource_trigger_action_test.go
```diff
@@ -55,6 +65,15 @@ func TestAccTriggerAction(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr("auth0_trigger_action.my_action_post_login", "trigger", "post-login"),
 					resource.TestCheckResourceAttrSet("auth0_trigger_action.my_action_post_login", "action_id"),
+					resource.TestCheckNoResourceAttr("auth0_trigger_action.my_action_post_login", "display_name"),
```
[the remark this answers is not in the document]
sergiught: Well we know that it will take the name of the action by default, so we can check for that one: `Test Action {{.testName}}`.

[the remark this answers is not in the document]
willvedd: That's in direct conflict with `TestCheckNoResourceAttr`.

internal/auth0/action/resource_trigger_action.go
willvedd: Can you be more specific? This is consistent across the entire codebase.

sergiught: It's consistent across the entire codebase for resources that get created and get deleted. The set of trigger bindings is static and is not something a customer can create or delete.

internal/auth0/action/resource_trigger_action.go
```diff
@@ -51,15 +52,22 @@ func NewTriggerActionResource() *schema.Resource {
 				ForceNew:    true,
 				Description: "The ID of the action to bind to the trigger.",
 			},
+			"display_name": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Computed:    true,
+				Description: "The name for this action within the trigger. This can be useful for distinguishing between multiple instances of the same action bound to a trigger.",
```
sergiught, line 59: Would be great to specify that this defaults to the action name if not set.

internal/auth0/action/resource_trigger_action.go
```diff
@@ -86,29 +94,80 @@ func createTriggerAction(_ context.Context, d *schema.ResourceData, m interface{
...
 	}
 
-	action, err := api.Action.Read(actionID)
-	if err != nil {
-		if mErr, ok := err.(management.Error); ok && mErr.Status() == http.StatusNotFound {
-			d.SetId("")
-			return nil
+	if displayName == "" {
+		action, err := api.Action.Read(actionID)
+
```
sergiught, line 99:
```suggestion
```

sergiught (approved):

PR #627, opened 2023-06-12
DXCDT-463: Mutex standardization

willvedd:
### 🔧 Changes

Mutexes are used to limit the damage from API requests between like-resources running in parallel and colliding, causing nefarious issues throughout the system. This is possible in two situations:
 - 1:1 relationship resources that operate on the same non-atomic endpoint by updating a big 
 - 1:many resources that overlap in functionality with deprecated "legacy" resources that manage relationships on the main resource directly

This PR was the result of an audit of all mutexes and attempts to standardize their usage.

General points: 
 - Ensure locking during create, update and delete (when appropriate)
 - Granular mutex keys (ex: `organizationID + "-connections"`) to prevent locking unlike resources
 - Ensure standard usage
 - Standardizing like-keys
 - Deletion of redundant calls

Additionally, it was considered that unlocking mutexes could be done synchronously, immediately after an API call. However, there is an added layer of safety when a resource's read function is also wrapped, and for little performance penalty too. 

### 📚 References

Original mutex PR: #425

### 🔬 Testing

All tests pass. And no major behavioral changes. Admittedly, the erroneous conditions that mutexes help solve are difficult to recreate, so no new tests.

### 📝 Checklist


internal/auth0/connection/resource_client.go
```diff
@@ -52,12 +52,12 @@ func NewClientResource() *schema.Resource {
...
 func createConnectionClient(ctx context.Context, data *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	api := meta.(*config.Config).GetAPI()
-	mutex := meta.(*config.Config).GetMutex()
 
 	connectionID := data.Get("connection_id").(string)
 
-	mutex.Lock(connectionID)
-	defer mutex.Unlock(connectionID)
+	mutex := meta.(*config.Config).GetMutex()
+	mutex.Lock(connectionID + "-clients") // Prevents colliding API requests between other `auth0_connection_client` resource.
```
willvedd, line 59: Appending the `-clients` string to the key makes it unique which makes it clear what other locks its competing with and also makes it unique so that other facets of the connection can be locked on.

sergiught (approved):

PR #647, opened 2023-06-20
DXCDT-442: `auth0_connection_database` resource

willvedd:
### 🔧 Changes

Introducing a new resource specific to the database (auth0) connection type.

 Currently, the sole `auth0_connection` resource encapsulates all 19 connection types that accounts for nearly 100 configurable fields, with the majority being specific to certain connection types. This is a DX problem because developers are left to their own devices to know which configurations apply to which connection type. In most cases, they'll need to sift through documentations and apply through trial-and-error until they reach the desired result.

Instead, this `auth0_connection_database` resource is the initial implementation whereby each connection type will be represented with its own resource. 

This PR not only implements the first connection type but sets a foundation for all other connection types to be implemented simliarly.

### 📚 References

Related Github issue: https://github.com/auth0/terraform-provider-auth0/issues/365

### 🔬 Testing

Retrofitting existing Auth0 connection tests to new resource.

### 📝 Checklist


docs/resources/connection_database.md
```diff
@@ -0,0 +1,187 @@
+---
+page_title: "Resource: auth0_connection_database"
+description: |-
+  With Auth0, you can define sources of users, otherwise known as connections, which may include identity providers (such as Google or LinkedIn), databases, or passwordless authentication methods. This resource allows you to configure and manage connections to be used with your clients and users.
+---
+
+# Resource: auth0_connection_database
+
+With Auth0, you can define sources of users, otherwise known as connections, which may include identity providers (such as Google or LinkedIn), databases, or passwordless authentication methods. This resource allows you to configure and manage connections to be used with your clients and users.
```
willvedd, line 9: Admittedly, this description makes a lot less sense being so generic, but I don't think it's time to abstract the description out further.

sergiught (approved):

PR #685, opened 2023-07-04
Remove connection metadata key validation

ErwinSteffens:
### 🔧 Changes

Validation on the `metadata` attribute of the `auth0_connection` resource now allows only keys with length 0-10 and the description shows a max of 10 metadata values. 

I have tested this with the API and you are allowed to:
* Add key lenghts longer then 255
* More then 10 metadata items 
* Values cannot be longer then 255 

I am not sure about the real limits, but it seems that the validation on the attribute does not match the validation on the API.

### 📚 References

-

### 🔬 Testing

Add connection with following metadata:

```
resource "auth0_connection" "my-connection" {
  name           = "my-connection"
  display_name   = "My Connection"

  metadata = {
    longer_then_10_chars = "value"
  }
}
```

### 📝 Checklist

### DIFF
--- a/docs/data-sources/connection.md
+++ b/docs/data-sources/connection.md
@@ -36,7 +36,7 @@ data "auth0_connection" "some-connection-by-id" {
 - `enabled_clients` (Set of String) IDs of the clients for which the connection is enabled.
 - `id` (String) The ID of this resource.
 - `is_domain_connection` (Boolean) Indicates whether the connection is domain level.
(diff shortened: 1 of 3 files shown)


sergiught (approved):
Thanks a lot for the contribution @ErwinSteffens 💯

ErwinSteffens:
> Thanks a lot for the contribution @ErwinSteffens 💯

No worries. Do you know when it will be released?

PR #709, opened 2023-07-12
Allowing GA version of node 18 actions runtime

willvedd:
### 🔧 Changes

As reported by #705 , a new `node18-actions` value exists specifically for Node 18 GA whereas the previous `node18` value is designated for beta. This PR adds that value to the validation list but also de-emphasizes the usage of the beta version to encourage users to leverage the GA version by default.

### 📚 References

Related Issue: #705 

### 🔬 Testing

Changing acceptance test to use new value, confirming its usage.

### 📝 Checklist


internal/auth0/action/resource.go
```diff
@@ -89,10 +89,11 @@ func NewResource() *schema.Resource {
 				ValidateFunc: validation.StringInSlice([]string{
 					"node12",
 					"node16",
-					"node18",
+					"node18",         // Node 18 beta.
+					"node18-actions", // Node 18 GA.
 				}, false),
 				Description: "The Node runtime. Defaults to `node12`. Possible values are: " +
-					"`node12`, `node16` or `node18`.",
+					"`node12`, `node16` or `node18-actions`.",
```
willvedd, line 96: While `node18` is technically a valid value, it will be removed at a future date and we should encourage our users to leverage the GA version.

sergiught (approved):

markus-winkler:
this problem still occurs in TerraformProvider 1.1.2 - can you please give an update how to use node18-actions as value with terraform ? Thx, Markus

PR #738, opened 2023-07-19
ci: Add `.semgrepignore` exclusion manifest

evansims:
This PR adds a `.semgrepignore` manifest file to exclude certain directories and documentation files from being incorrectly flagged by Semgrep. This will help avoid false positive warnings being triggered.

### DIFF
--- a/.semgrepignore
+++ b/.semgrepignore
@@ -1,4 +1,5 @@
 docs/
 examples/
 test/
-*.md
+\_.md
+internal/\*\_test.go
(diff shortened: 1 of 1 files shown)


evansims:
@willvedd Appears `actions/setup-go` isn't picking up on 1.20.6 from the [versions manifest](https://github.com/actions/go-versions/blob/main/versions-manifest.json) quite yet, which resolves [CVE-2023-29406](https://github.com/golang/go/issues/60374) — which is what Snyk is warning about in the failing CI. Not sure how long it caches the manifest, but it should pick up on it soon on it's own.

willvedd (approved):
Thanks for the proactive change!

PR #779, opened 2023-08-14
GH-273: Increasing number of retries

willvedd:
### 🔧 Changes

Version 1 of the Go SDK introduced a new way of expressing retry logic when instantiating the management client. This change is a deviation from v0 which implemented unlimited retries by default. Even with three retries, users are [still experiencing 429 errors in their workflows](https://github.com/auth0/terraform-provider-auth0/issues/273#issuecomment-1668423627) so this PR increases the number of retries to 12. This number was selected arbitrarily but high enough to smooth over most instances without overwhelming our servers in the case of 500s.

### 📚 References

- [Related Github Issue](https://github.com/auth0/terraform-provider-auth0/issues/273)

### 🔬 Testing

Tested on the Go SDK-side. _Should_ make our acceptance tests more reliable too.

### 📝 Checklist

### DIFF
--- a/internal/acctest/acctest.go
+++ b/internal/acctest/acctest.go
@@ -66,7 +66,7 @@ func configureTestProviderWithHTTPRecordings(httpRecorder *recorder.Recorder) sc
 			management.WithStaticToken("insecure"),
 			management.WithClient(httpRecorder.GetDefaultClient()),
 			management.WithDebug(debug),
-			management.WithRetries(3, []int{http.StatusTooManyRequests, http.StatusInternalServerError}),
+			management.WithRetries(12, []int{http.StatusTooManyRequests, http.StatusInternalServerError}),
 		}
 
 		if domain != RecordingsDomain {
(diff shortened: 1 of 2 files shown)


ewanharris (approved):

luislew:
Question: why arbitrary number of retries instead of using the `x-ratelimit-reset` header to determine how long to wait until the rate limit resets and _then_ retrying? As an example, Datadog's TF provider does this: https://github.com/DataDog/terraform-provider-datadog/blob/103da01a8bcadd3de35849909cfcc072da6a70b5/datadog/internal/transport/custom_transport.go#L89-L96

ioquatix:
Won't increasing the number of retries, without any back-off logic (delay/sleep) make overloading worse?

yourinium:
Any updates on solving the 429 issue?

ewanharris:
To clarify the behaviour of the Go SDK retry strategy in v1:

* A delay is calculated using exponential backoff with jitter, this is capped at 10 seconds
* The `x-ratelimit-reset` header is consulted and if the time to the reset is longer than the calculated delay (but lower than 10 seconds) we will use that. There is a cap here because the rate limit is not consistent across all management APIs.

So increasing the retries should help with reducing this issue as it will lead to an increase in delay between requests.

PR #780, opened 2023-08-14
Mark error_page elements as optional

jeff-auth0:
### 🔧 Changes

As Per discussion on slack [here](https://okta.slack.com/archives/C02SE6ZBGVC/p1691990805452079), currently all `error_page` elements are marked as `required=true` whereas in Management API [ref](https://auth0.com/docs/api/management/v2/tenants/patch-settings) these elements are `optional`. 

In this pull request, `error_page` elements are marked as `false` and it is aligned with Management [API](https://auth0.com/docs/api/management/v2/tenants/patch-settings).

### 📝 Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ override.tf.json
 .vscode/*
 !.vscode/tasks.json
 *.code-workspace
+.history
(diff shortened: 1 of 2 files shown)


willvedd (changes requested):
Removing the `Required` and/or setting these properties to optional is not what we want because it would be possible to provide a block with neither properties. We want _some_ validation here after all. Instead, `AtLeastOneOf` should be leveraged to enforce that at least one of the properties is set. See https://github.com/auth0/terraform-provider-auth0/blob/c6e2185488446464a5e3b79a62032f2440ed1e44/internal/auth0/client/resource.go#L596 for reference.

Additionally, I ask you base this off the `v1` branch because v0.x is still based off `main` and we are trying to eliminate any additional releases.

PR #866, opened 2023-10-19
Fix pkce connection docs typo

Rigdon:
### 🔧 Changes

This was specified as the value `auth0` but, according to the api, the actual correct value is `auto`, as shown in the examples as well.

### 📝 Checklist

- [X] All new/changed/fixed functionality is covered by tests (or N/A) N/A

### DIFF
--- a/docs/resources/connection.md
+++ b/docs/resources/connection.md
@@ -767,7 +767,7 @@ Optional:
 
 Required:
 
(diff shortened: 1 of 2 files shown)


sergiught (approved):
Thanks for the contribution @Rigdon !

PR #946, opened 2024-04-10
Detect organization membership by roles instead of scanning the member list

### LINKED ISSUE #925: Can't import or create `auth0_organization_member` on Organizations with more than 50 members
### Checklist

- [X] I have looked into the [README](https://github.com/auth0/terraform-provider-auth0#readme) and have not found a suitable solution or answer.
- [X] I have looked into the [documentation](https://registry.terraform.io/providers/auth0/auth0/latest/docs) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/terraform-provider-auth0/issues) and have not found a suitable solution or answer.
- [X] I have upgraded to the [latest version](https://github.com/auth0/terraform-provider-auth0/releases/latest) of this provider and the issue still persists.
- [X] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

I have an Organization with many members that I am currently trying to associate new members using the `auth0_organization_member`. When I try to add new members, I get the "inconsistent result after apply" error with

```
│ "provider[\"registry.terraform.io/auth0/auth0\"]" produced an unexpected new value: Root resource was present, but now absent.
```

This only happens when there are at least 50 members in the organization. I expect that this correlates with the default page size for the Go SDK. Currently, there is only one query to find organization members, and existence is checked by a simple loop over the results:

https://github.com/auth0/terraform-provider-auth0/blob/21e8c526ed1c7bfd3d889daef4efa5069f66a833/internal/auth0/organization/resource_member.go#L61-L71

### Expectation

Terraform should be able to add `auth0_organization_member`s to Organizations of arbitrary size. Rather than checking for an ID in every page of the member list, I think it would be better to check that the `GET  /api/v2/organizations/{id}/members/{user_id}/roles` endpoint doesn't return a 404 response. Alternatively, the `organization_id:` field could be used for a Lucene search on the user search endpoint.

### Reproduction

1. Create an Organization with at least 50 members. I believe this is the only requirement since Auth0 seems to order members in order of insertion. It might be necessary to go beyond this - my organization has around 190 members currently.
2. add a new `auth0_organization_member` resource
   ```hcl
    resource "auth0_organization_member" "bulk_test_users" {
      count = length(auth0_user.bulk_test_users)
    
      organization_id = "{id}"
      user_id         = auth0_user.bulk_test_users[count.index].user_id
    }
   ```

1.2.0

### Terraform version

1.3.7

kherock:
Fixes #925. The deprecation and removal of the `roles` field for version 1.0.0 came with a bug that prevented organization members from being added to organizations with 50 or more members.

The current code only finds members if they appear in the first 50 members of an organization. This corresponds to the default page size of 50 in the Auth0 Go SDK.

### 🔧 Changes

Rather than scanning through many pages of members, this PR restores some of the old behavior of checking whether the user has any roles associated with their organization membership. If Auth0 returns a 404 on this endpoint, it means that the user is not a member.

### 🔬 Testing

I've tested these changes in my own fork and they work as expected - currently I do not have much time to dedicate on adding tests. I hope that due to the nature of this bug, this fix should be acceptable as long as it doesn't cause any regressions 🤞😬

### 📝 Checklist

### DIFF
--- a/internal/auth0/organization/resource_member.go
+++ b/internal/auth0/organization/resource_member.go
@@ -57,20 +57,12 @@ func readOrganizationMember(ctx context.Context, data *schema.ResourceData, meta
 	api := meta.(*config.Config).GetAPI()
 
 	organizationID := data.Get("organization_id").(string)
+	userID := data.Get("user_id").(string)
 
-	members, err := api.Organization.Members(ctx, organizationID)
+	_, err := api.Organization.MemberRoles(ctx, organizationID, userID)
 	if err != nil {
 		return diag.FromErr(internalError.HandleAPIError(data, err))
 	}
-
-	userID := data.Get("user_id").(string)
-	for _, member := range members.Members {
-		if member.GetUserID() == userID {
-			return nil
-		}
-	}
-
-	data.SetId("")
 	return nil
 }
 


kherock:
Thanks for looking at this -
I'm not sure if I agree with this approach being indirect, roles are an intrinsic property of every member. It also performs significantly better for projects that only declare one or two members in a large organization.

The approach in the other PR does not perform well with organizations containing many members - I would only use that for a Terraform  resource maintaining an exhaustive list of all members.

E.g. of an organization has 10 pages of members before any Terraform-managed members are added, those 10 pages will have to be fetched for every single member added!

kherock:
@developerkunal the implemented solution is actually causing us to instantly hit the Auth0 global rate limit and throttle management API requests in our production environments. One of the organizations that were adding users to has over 20 pages!

Can this implementation still be considered please?

developerkunal:
Hi @kherock,

I hope you're having a good day!

I wanted to thank you for your input; your contribution means a lot to us. I understand your point, but we want to make sure we're using the proper members endpoint instead of just depending on member roles to determine if a user is part of an organization. If you have any questions or concerns, please don't hesitate to ask.

Thanks again!

Issue #1030, opened 2024-09-20
Add support to Backchannel logout initiators


ncastrocosta:
### Checklist

- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

Add support to allow configure the 'Backchannel logout initiators' of an application client.

![image](https://github.com/user-attachments/assets/d07bc959-a572-45f2-9d7e-841aa3e62a34)

### Describe the ideal solution

Add the following properties should be added to the `auth0_client` resource:
- `oidc_backchannel_logout_initiators`: list of strings. Default: ["idp-logout", "rp-logout"]
- `oidc_backchannel_logout_initiators_mode`: string (all or custom). Default: custom

If the user chooses for `oidc_backchannel_logout_initiators_mode = "all"`, the `oidc_backchannel_logout_initiators` must not be present.

This is the behavior I got from Auth API.

duedares-rvj:
This is being supported as part of [1045](https://github.com/auth0/terraform-provider-auth0/pull/1045)

Thanks!

PR #1045, opened 2024-10-08
Add OIDC backchannel initiators

bryanroute:
### 🔧 Changes

This adds two new properties to the auth0_client resource:

oidc_backchannel_logout_initiators_mode to set the mode for which initiators should get a logout token. Valid values are "all" or "custom"

oidc_backchannel_logout_initiators to set which specific initiators should get a logout token if the mode is set to "custom"

This allows all the configuration around the OIDC backchannel logout to be set via terraform where as currently you can only set the URL and nothing else. 

Worth calling out this is my first PR working with a terraform provider so if I'm missing something about the terraform SDK and how it is supposed to work I apologize in advance. I just really want this functionality added and this seems to do it. 

### 📚 References

Addresses this open issue with the proposed design
https://github.com/auth0/terraform-provider-auth0/issues/1030 

Also addresses this support ticket: https://support.auth0.com/tickets/02368461 

### 🔬 Testing

Test checks were added to the existing tests however http recordings need to be regenerated for the tests to pass and it appears from what I could figure that has to be done by someone that has access to the test tenant ```terraform-provider-auth0-dev.eu.auth0.com```

I did run the e2e tests against my own test tenant and everything passed except for some tests related to features not enabled in my test tenant.

This was manually tested and verified by creating a client with none of the back channel properties set, one with the mode all and one with the mode custom. Everything appears to function as I expected.

I do have an open question whether the approach to handling the default mode/initiators is the correct one however I'm not sure of a better approach and the behavior seems to mirror similar behavior I've observed with other terraform providers I've worked with. 

### 📝 Checklist


internal/auth0/client/expand.go
```diff
@@ -41,7 +41,7 @@ func expandClient(data *schema.ResourceData) (*management.Client, error) {
 		InitiateLoginURI:                   value.String(config.GetAttr("initiate_login_uri")),
 		EncryptionKey:                      value.MapOfStrings(config.GetAttr("encryption_key")),
 		IsTokenEndpointIPHeaderTrusted:     value.Bool(config.GetAttr("is_token_endpoint_ip_header_trusted")),
-		OIDCBackchannelLogout:              expandOIDCBackchannelLogout(data),
```
developerkunal, line 44: Instead of removing deprecate it. it's not going to be a major release.

internal/auth0/client/flatten.go
```diff
@@ -557,7 +557,9 @@ func flattenClient(data *schema.ResourceData, client *management.Client) error {
 		data.Set("initiate_login_uri", client.GetInitiateLoginURI()),
 		data.Set("signing_keys", client.SigningKeys),
 		data.Set("client_metadata", client.GetClientMetadata()),
-		data.Set("oidc_backchannel_logout_urls", client.GetOIDCBackchannelLogout().GetBackChannelLogoutURLs()),
```
developerkunal, line 560: Same here don't remove this just deprecate this.

internal/auth0/client/resource.go
```diff
@@ -116,6 +116,28 @@ func NewResource() *schema.Resource {
...
+					Type: schema.TypeString,
+					ValidateFunc: validation.StringInSlice([]string{
+						"rp-logout", "idp-logout", "password-changed", "session-expired", "session-revoked",
+						"account-deleted", "email-identifier-changed", "mfa-phone-unenrolled", "account-deactivated",
+					}, false),
+				},
+				Optional: true,
+				Description: "Which specific initiators should receive a logout token. Possible values are " +
+					"rp-logout, idp-logout, password-changed, session-expired, session-revoked, account-deleted, email-identifier-changed, mfa-phone-unenrolled, account-deactivated",
+			},
```
developerkunal, line 140: Perhaps it would be better to create a nested `OIDCLogout` structure instead of the current approach, as `oidc_backchannel_logout_urls` already contains other values. We can deprecate `oidc_backchannel_logout_urls` and introduce `OIDCLogout`, with nested properties such as `oidc_backchannel_logout_urls` and `BackChannelLogoutInitiators`, similar to the implementation in the SDK.

developerkunal (changes requested):

internal/auth0/client/expand.go
bryanroute: The concern I had with that approach is you introduce the potential of setting the back channel logout urls in two different places, via the depreciated property and via the new nested properties. I imagine the Auth0 API will reject if you try that but I wasn't sure?

I also was trying to follow the proposal in this issue https://github.com/auth0/terraform-provider-auth0/issues/1030
Generally I do prefer flat json to nested properties. 

It seems the main reason for this change is your comment about backwards compatibility. These properties exist on the API and are set even though you can't control them via terraform currently. I guess the concern is someone gets a provider update and then it will try to set them back to the default values if they don't add their current values to their terraform? I can see that would be an issue. I just hate resorting to nested JSON but I suppose there isn't a better option.

bryanroute:
@developerkunal I just updated with your suggestion of using a oidc_logout block with nested attributes. There are still some failing tests when I run E2E in my test tenant not related to features we don't have enabled. 

The problem appears to be that the API returns the logout urls both in the OIDCBackchannelLogout object and the OIDCLogout object. I've tried a couple different approaches but rolled them back because they all had unintended side effects. The problem is we don't necessarily know which approach in the flatten to take, ignore the old setting or ignore the new one. The paths I've gone down have all caused another test to fail. I'm going to step away and then come back at it fresh. If you have any suggestions for handling it I'd appreciate them.

Seha16:
> @developerkunal I just updated with your suggestion of using a oidc_logout block with nested attributes. There are still some failing tests when I run E2E in my test tenant not related to features we don't have enabled.
> 
> The problem appears to be that the API returns the logout urls both in the OIDCBackchannelLogout object and the OIDCLogout object. I've tried a couple different approaches but rolled them back because they all had unintended side effects. The problem is we don't necessarily know which approach in the flatten to take, ignore the old setting or ignore the new one. The paths I've gone down have all caused another test to fail. I'm going to step away and then come back at it fresh. If you have any suggestions for handling it I'd appreciate them.

@bryanroute If I understand you correctly, I think, it's better to ignore old setting `oidc_backchannel_logout_urls` because if there is introduced a new block `oidc_logout`, it requires `backchannel_logout_urls` in it. Thus it guarantees that Backchannel Logout URL will be present in the execution plan.

docs/data-sources/client.md
```diff
@@ -62,6 +62,7 @@ data "auth0_client" "some-client-by-id" {
 - `native_social_login` (List of Object) Configuration settings to toggle native social login for mobile native applications. Once this is set it must stay set, with both resources set to `false` in order to change the `app_type`. (see [below for nested schema](#nestedatt--native_social_login))
 - `oidc_backchannel_logout_urls` (Set of String) Set of URLs that are valid to call back from Auth0 for OIDC backchannel logout. Currently only one URL is allowed.
```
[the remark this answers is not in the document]
bryanroute: Do we want to change the data source to pull back the data differently? I didn't change the data source to add the new nested structure and this is auto generated from that so that is why it isn't marked as depreciated.

bryanroute:
@Seha16 I'm not sure I understand your suggestion. If I don't set the old oidc_backchannel_logout_urls then won't that generate a unexpected plan diff for someone using the old property and not implementing the new block?

I'm not sure if I'm understanding your suggestion.

Seha16:
@bryanroute I just wanted to reply on your message

> The problem is we don't necessarily know which approach in the flatten to take, ignore the old setting or ignore the new one.

And I want to say that I like you current implementation and agree with you to ignore `oidc_backchannel_logout_urls` if `oidc_logout` is introduced:

```go
func flattenOIDCBackchannelURLs(backchannelLogout *management.OIDCBackchannelLogout, logout *management.OIDCLogout) []string {
  if logout != nil {
    return nil
  } else {
    return backchannelLogout.GetBackChannelLogoutURLs()
  }
}

...

data.Set("oidc_backchannel_logout_urls", flattenOIDCBackchannelURLs(client.GetOIDCBackchannelLogout(), client.GetOIDCLogout())),
data.Set("oidc_logout", flattenOIDCLogout(client.GetOIDCLogout())),
```

There was silence in this PR and no replies on your comments for a while. So I thought voting for your solution may bring some attention of the project contributors.

docs/data-sources/client.md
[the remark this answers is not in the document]
Seha16: yeah, make sense, for client retrieval it is probably not deprecated.

bryanroute:
Ah @Seha16 I thought you were from Auth0. I got a message on my support ticket tracking the missing functionality. They are supposed to be reviewing my PR soon and figured that you were the one doing so. Appreciate drawing attention to it. I'm hoping I hear back from the team soon so we can get this functionality.

developerkunal:
Hi @bryanroute ,
we need the commits to be signed. can you please sign the commits.

duedares-rvj (approved):

developerkunal (approved):

PR #1132, opened 2025-01-28
node22 support

pashaosipyants:
Can we please 🙏 have at least this hotfix?
And you'll update the docs, etc later.
This validation is really blocking


duedares-rvj:
Please bear with us. We already have plans to release this change as part of another PR shortly. 
Thanks!

PR #1145, opened 2025-02-04
Added flatten/expand for token_exchange

duedares-rvj:
1. `custom-token-exchange` trigger does not support node18-actions as runtime. It only supports node18 and node22. With the present terraform configuration which manipulated node18 input to node18-actions, it was practically impossible to create an action for this trigger with node18 as runtime. A required clause was added to support this and maintain the same functionality for other triggers.
2. Missing expand/flatten rules where added for auth0_client `token_exchange` param

### 📝 Checklist


ramya18101 (approved):

developerkunal (approved):
LGTM

PR #1355, opened 2025-09-11
feat: Support Token Vault

kushalshit27:
### 🔧 Changes

Key Features:

- Added resource_server as a valid `app_type` for clients
- Introduced `resource_server_identifier` field for clients to link them to specific resource servers
- Added `client_id` field to resource servers to show linked client information

### Example:
````hcl
resource "auth0_resource_server" "my_resource_server" {
  name       = "Example API"
  identifier = "https://api.example.com"
}

  name                       = "Resource Server Client"
  app_type                   = "resource_server"
  resource_server_identifier = auth0_resource_server.my_resource_server.identifier
  # ..........
}

data "auth0_resource_server" "example" {
  resource_server_id = auth0_resource_server.my_resource_server.id
}

# Output the linked client ID
output "linked_client_id" {
  value = data.auth0_resource_server.example.client_id
}

````

### 🔬 Testing
- Unit test added
- E2E recordings updated

### 📝 Checklist

### DIFF
--- a/internal/auth0/client/expand.go
+++ b/internal/auth0/client/expand.go
@@ -57,7 +57,6 @@ func expandClient(data *schema.ResourceData) (*management.Client, error) {
 		SessionTransfer:                    expandSessionTransfer(data),
 		ComplianceLevel:                    value.String(config.GetAttr("compliance_level")),
 		TokenQuota:                         commons.ExpandTokenQuota(config.GetAttr("token_quota")),
-		ResourceServerIdentifier:           value.String(config.GetAttr("resource_server_identifier")),
 	}
 
 	if data.IsNewResource() && client.IsTokenEndpointIPHeaderTrusted != nil {
@@ -71,6 +70,10 @@ func expandClient(data *schema.ResourceData) (*management.Client, error) {
 		case "regular_web", "non_interactive":
 			client.TokenEndpointAuthMethod = auth0.String("client_secret_post")
 		}
+	}
+
+	if data.IsNewResource() && client.ResourceServerIdentifier != nil {
+		client.ResourceServerIdentifier = value.String(config.GetAttr("resource_server_identifier"))
 	}
 
 	defaultConfig := data.GetRawConfig().GetAttr("default_organization")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: internal/auth0/resourceserver/resource.go [hunk 1]
```diff
@@ -303,6 +303,13 @@ func NewResource() *schema.Resource {
 					},
 				},
 			},
+			"client_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Optional: true,
```
duedares-rvj, line 309: If this field cannot be set, and is always Computed, we should consider removing `Optional:true` 
And if it can be set, then we have to add code to resourceserver/expand.go 

I believe former is the case here.

internal/auth0/client/expand.go
```diff
@@ -57,6 +57,7 @@ func expandClient(data *schema.ResourceData) (*management.Client, error) {
 		SessionTransfer:                    expandSessionTransfer(data),
 		ComplianceLevel:                    value.String(config.GetAttr("compliance_level")),
 		TokenQuota:                         commons.ExpandTokenQuota(config.GetAttr("token_quota")),
+		ResourceServerIdentifier:           value.String(config.GetAttr("resource_server_identifier")),
```
duedares-rvj, line 60: What does the API return if you try to change this value?
If it returns a valid error, then we are good.

Else, we might have to include this under:

```go
if data.IsNewResource() {
		client. ResourceServerIdentifier = value.String(config.GetAttr("resource_server_identifier"))
	}
```

And ignore this field for put/patch calls.

internal/auth0/resourceserver/resource.go
duedares-rvj: Consider removing `Optional:true` in that case. Just Computed.

internal/auth0/client/expand.go
duedares-rvj: No. But it will avoid sending the field in an update call. (since we don't support any PATCH, we have to avoid sending it too)

But, since it's called out on description `This property CANNOT be changed, once the client is created.` We should be good. You can also add explicit check like:

```go
if !data.IsNewResource() && data.HasChange("resource_server_identifier") {
		return nil, fmt.Errorf("your error")
	}
```

PR #1436, opened 2025-12-15
Add write-only support for event streams

### LINKED ISSUE #1402: Add support for write-only token in `auth0_event_stream`
### Checklist

- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

It would be great if the `auth0_event_stream` resource would support a [write-only](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/write-only) token argument when setting it up with a webhook without compromising the security of the token value and storing it in the terraform state. 

### Describe the ideal solution

Here's what I believe an example could look like for a token to be passed from outside terraform (because using another provider's ephermeral resource runs into all kinds of weirdness with unsupported provider to provider dependency)

```hcl
# this variable can be passed using command line arguments to terraform
variable "webhook_token" {
  description = "The webhook token"
  type        = string
  sensitive   = true
}

# Creates an event stream of type webhook
resource "auth0_event_stream" "my_event_stream_webhook" {
  name             = "my-webhook"
  destination_type = "webhook"
  subscriptions = [
    "user.created",
    "user.updated"
  ]

  webhook_configuration {
    webhook_endpoint = "https://eof28wtn4v4506o.m.pipedream.net"

    webhook_authorization {
      method = "bearer"
      # read the secret from a sensitive variable
      token_wo  = var.webjook_token
      token_wo_version  = "1"
    }
  }
}
```

### Alternatives and current workarounds

There are currently no alternatives **with terraform**.

The only other option I can think of is just not terraforming it

### Additional context

_No response_

siddiqui-io:
### 🔧 Changes

1. Add write only support for event streams
2. Update the expand and flatten logic

### 📚 References

Closes #1402 

### 🔬 Testing

1. Added unit tests for the new logic
2. Tested manually by creating and updating event-streams using local builds

### 📝 Checklist

### DIFF
--- a/internal/auth0/eventstream/resource_test.go
+++ b/internal/auth0/eventstream/resource_test.go
@@ -154,7 +154,7 @@ func TestAccEventStream(t *testing.T) {
 					resource.TestCheckResourceAttr("auth0_event_stream.my_event_stream_webhook", "webhook_configuration.0.webhook_endpoint", "https://test.com"),
 					resource.TestCheckResourceAttr("auth0_event_stream.my_event_stream_webhook", "webhook_configuration.0.webhook_authorization.0.method", "bearer"),
 
-					resource.TestCheckNoResourceAttr("auth0_event_stream.my_event_stream_webhook", "webhook_configuration.0.webhook_authorization.0.token_wo"), // token_wo is write-only
+					resource.TestCheckNoResourceAttr("auth0_event_stream.my_event_stream_webhook", "webhook_configuration.0.webhook_authorization.0.token_wo"),
 				),
 			},
 		},


internal/auth0/eventstream/expand.go
```diff
@@ -44,36 +44,52 @@ func expandEventStreamDestination(data *schema.ResourceData) *management.EventSt
...
-				}
-
-				if method == "basic" {
-					if v, ok := auth["username"].(string); ok {
-						authMap["username"] = v
-					}
-					if v, ok := auth["password"].(string); ok {
-						authMap["password"] = v
+		rawConfig := data.GetRawConfig()
+		if !rawConfig.IsNull() {
```
bkiran6398: Multiple nested if statements can be reduced by using early return/break statements for sad paths to improve readability.
eg:
```go
if rawConfig.IsNull() {
    break
}
// other logic
if webhookConfigRaw.IsNull() || webhookConfigRaw.LengthInt() == 0 {
    break
}
// continue happy path
```

internal/auth0/eventstream/flatten.go
```diff
@@ -78,17 +78,36 @@ func flattenEventStreamDestination(data *schema.ResourceData, dest *management.E
...
 				if t := data.Get("webhook_configuration.0.webhook_authorization.0.token"); t != nil {
 					authMap["token"] = t
 				}
+
+				// The token_wo is write-only and should NOT be read from API or stored in state.
+				// Instead, we only preserve the version from config.
+				// So no action WRT token_wo here.
+			}
+
+			// The token_wo_version is stored in state to track changes.
```
bkiran6398: - To be moved inside `else` block above.
- Would it require to explicitly set `password_wo_version` to `0` in case of `bearer` auth method.

PR #1522, opened 2026-04-08
Deprecate phone MFA configuration in auth0_guardian if PhoneConsolidatedExperience is enabled

bkiran6398:
### 🔧 Changes

Deprecates the `phone` block in the `auth0_guardian` resource in favor of the `auth0_phone_provider` resource for tenants with Unified Phone Experience (UPE) enabled.

- Marks the `phone` block as deprecated with guidance to use `auth0_phone_provider` instead.
- Skips reading phone configuration via deprecated Guardian APIs when `phone_consolidated_experience` is enabled on the tenant.
- Conditionally disables phone MFA on resource deletion only when the tenant has **not** migrated to UPE, preventing conflicts with the new `auth0_phone_provider` resource.

### 🔬 Testing

- Verified that tenants with `phone_consolidated_experience` enabled skip deprecated Guardian phone API calls during read and delete.
- Verified that tenants without UPE continue to use the existing phone MFA flow unchanged.
- Confirmed deprecation warning surfaces in `terraform plan` output when the `phone` block is present.

### 📝 Checklist


bkiran6398:
Raised a new PR #1596 to handle the same.

PR #1534, opened 2026-04-19
fix(client-grant): respect subject_type when adopting existing grants

asafdav:
## Summary

`createClientGrant` in `internal/auth0/client/resource_grant.go` currently lists grants by `(client_id, audience)` only, and when the list is non-empty it adopts `grantList.ClientGrants[0]` unconditionally — ignoring `subject_type`. Auth0's data model supports multiple client grants per `(client_id, audience)` differentiated by `subject_type` (one for `"client"` access, one for `"user"` access — see `go-auth0/management/client_grants.go`, where `SubjectType` is a per-grant scalar, not an array). The Auth0 Dashboard itself creates independent grants when the User Access / Client Access toggles are flipped.

Today, configuring two `auth0_client_grant` resources targeting the same client + audience with different `subject_type` values silently collapses them onto a single grant id. Terraform then reports no drift, but the second grant Auth0 should have created was never provisioned.

## Reproduction

The following configuration reproduces the bug against v1.43.0:

```hcl
resource "auth0_client" "my_app" {
  name = "example"
}

resource "auth0_resource_server" "my_api" {
  name       = "example"
  identifier = "https://example.api/"
  subject_type_authorization {
    user   { policy = "allow_all" }
    client { policy = "require_client_grant" }
  }
}

resource "auth0_client_grant" "client_access" {
  client_id    = auth0_client.my_app.id
  audience     = auth0_resource_server.my_api.identifier
  subject_type = "client"
  scopes       = ["read:foo"]
}

resource "auth0_client_grant" "user_access" {
  client_id    = auth0_client.my_app.id
  audience     = auth0_resource_server.my_api.identifier
  subject_type = "user"
  scopes       = ["read:foo"]
}
```

Observed: both resources end up with the same Auth0 grant id (whichever was created first); the other grant is never created. The behavior matches the adoption path in `createClientGrant` (lines ~115-118 in v1.43.0).

## Fix

Walk `grantList.ClientGrants` and adopt an existing grant **only** if its `subject_type` matches the user's configuration. When empty, normalize to `"client"` (the default documented on the `subject_type` schema field in the same file). If no match is found, proceed to `expandClientGrant` + `api.ClientGrant.Create` so Auth0 creates a second grant differentiated by `subject_type` — matching what the Dashboard does.

A small helper `normalizeSubjectType` applies the empty → `"client"` normalization on both sides of the comparison so that a configuration with `subject_type = "client"` correctly matches a pre-existing grant whose API payload omits or empties `subject_type`.

## Testing

- Added a focused unit test `TestNormalizeSubjectType` in `internal/auth0/client/resource_grant_internal_test.go` covering the helper that underpins the adoption comparison.
- `go build ./...` and `go vet ./...` pass locally.
- `go test -run TestNormalizeSubjectType ./internal/auth0/client/...` passes.
- Full acceptance tests were **not** run in this branch because they require a live Auth0 tenant; happy to run them in CI.

A provider-level mock for `createClientGrant`'s adoption flow was considered but is out of scope for a targeted bug fix — the existing `resource_grant_test.go` is entirely acceptance-based.

## Notes

Happy to iterate on test coverage or change style (e.g., exporting the helper, inlining the normalization, matching an existing pattern in the repo).


bkiran6398:
Hi @asafdav, Thank you so much for your contribution, and apologies for the delayed review.

The implementation looks good to me overall. I just have a couple of small suggestions:
* Would it be possible to make the comments a bit shorter and more concise.
* I think it would also be helpful to introduce a shared constant (e.g. `defaultSubjectType`) and use it in both the docs and `normalizeSubjectType` to have a single source of truth.

Also, Could you please GPG-sign your commits? It's a requirement before we can merge into `master`. This [guide](https://stackoverflow.com/questions/41882919/is-there-a-way-to-gpg-sign-all-previous-commits) can help you.

Thanks again for taking the time to contribute!

bkiran6398 (approved):

ShawnFumo:
Just a quick thanks for doing this! We actually JUST ran into this today (we wanted to restrict the user-delegation grants to just specific apps, and so needed to add the user grant in addition to the existing m2m grant at the same time we enabled the restricted mode). At first it seemed like it would work (diff showed adding the new grant), but it just changed the existing. Luckily we caught it fairly early, but wouldn't have been able to go forward without the fix. Now we will once we get a chance to update to the latest provider.

PR #1541, opened 2026-04-23
feat: add write-only support for auth0_action secrets

Asarew:
### Changes

Adds write-only support for `auth0_action` secrets so values no longer end up in Terraform state. `auth0_action.secrets.value` is currently marked `Sensitive` but is still persisted to `.tfstate` in plain text. This follows the same `_wo` / `_wo_version` pattern added for `auth0_event_stream` webhook credentials in v1.39.0.

New attributes on `auth0_action`:

- `secrets_wo` (Block List) — `{ name, value }` entries, `value` is `WriteOnly` + `Sensitive` and never stored in state.
- `secrets_wo_version` (Number) — bump this to push new secret values to the API.

`secrets_wo` and `secrets_wo_version` are `RequiredWith` each other, and `secrets_wo` conflicts with the existing `secrets` block.

### Why a new `TypeList` instead of `WriteOnly` inside the existing `TypeSet`

`TypeSet` hashes elements from all attributes. A `WriteOnly` value is present in config but not in state, so the set-element hash never matches and you get perpetual diffs. Using a `TypeList` for `secrets_wo` avoids this since elements are addressed by index. Same approach as `password_wo` / `token_wo` inside `webhook_authorization`.

### Usage

```hcl
variable "action_api_key" {
  type      = string
  sensitive = true
  ephemeral = true
}

resource "auth0_action" "my_action" {
  name    = "Post Login Action"
  runtime = "node22"
  deploy  = true
  code    = file("action.js")

  supported_triggers {
    id      = "post-login"
    version = "v3"
  }

  secrets_wo {
    name  = "API_KEY"
    value = var.action_api_key
  }

  secrets_wo_version = 1
}
```

Rotating a secret = update the variable and bump `secrets_wo_version`.

### 📚 References

- Write-only args on `auth0_event_stream` (v1.39.0) — same pattern.
- https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments

### 🔬 Testing

- `go vet ./...`, `go build ./...`, `go test ./internal/auth0/action/...` pass locally.
- Docs regenerated with `go generate`.

Acceptance tests added (require a tenant, not run locally):

- `TestAccActionWithSecretsWO` — create with `secrets_wo_version = 1`, assert `secrets_wo.0.value` is absent from state, then bump to `2` with a new value.
- `TestAccActionSecretsWOConflictsWithSecrets` — expects the conflict error.
- `TestAccActionSecretsWORequiresVersion` — expects the missing-`secrets_wo_version` error.

Unit test `TestCheckForUntrackedActionSecretsWithSecretsWO` covers the unmanaged-secrets guard on `secrets_wo` input.

### 📝 Checklist


Asarew:
@sergiught @ramya18101 would you guys maybe be able to help out with moving this PR forward?

bkiran6398:
Thanks a lot for this contribution @Asarew.

The write-only approach and the `TypeList`-over-`TypeSet` rationale are spot on. I tested it end-to-end against a live tenant and hit one limitation around removing/clearing `secrets_wo` entries. Everything else (create, value rotation via `secrets_wo_version`, conflict/required validation) works correctly.

### Repro
Create an action with 3 `secrets_wo` entries, then remove two of them from config and `apply` **without** bumping `secrets_wo_version`:

- `terraform plan` shows the two blocks being removed, and `apply` reports success.
- But the `PATCH` body sent to the API contains **no `secrets` field**, so the API silently keeps all 3 secrets. They're now **orphaned live on the API**, still usable by the action, but invisible to Terraform. The next `plan` says "No changes."

### Root cause
Both the expand step and the `preventErasingUnmanagedSecrets` guard are gated solely on `secrets_wo_version`. But the secret **`names`** are tracked in state (only `value` is write-only), so a structural change to `secrets_wo` is fully observable via `HasChange("secrets_wo")`, it just isn't being checked. As a result:

- **Removing without a version bump** → silent orphan + state drift (above).
- **Removing *with* a version bump** → the guard compares the API's names (e.g. `foo`, `bar`) against the *new config only* (`foo`) and aborts with a false `"Unmanaged Action Secret"` error, blocking the removal.

So removing a `secrets_wo` entry is effectively impossible today. The fix is to make `secrets_wo` symmetric with how the existing `secrets` block is handled: trigger on the structural change and aggregate old+new names. Suggestions inline.

internal/auth0/action/expand.go
```diff
@@ -30,6 +30,11 @@ func expandAction(data *schema.ResourceData) *management.Action {
 
 	if data.HasChange("secrets") {
 		action.Secrets = expandActionSecrets(config.GetAttr("secrets"))
+	} else if data.IsNewResource() || data.HasChange("secrets_wo_version") {
+		secretsWO := config.GetAttr("secrets_wo")
+		if !secretsWO.IsNull() && secretsWO.LengthInt() > 0 {
+			action.Secrets = expandActionSecretsWO(secretsWO)
+		}
 	}
```
bkiran6398: `secrets_wo` names live in state, so trigger the expand on `HasChange("secrets_wo")` (catches add/rename/remove), not just the version. Also handle the "all entries removed" case by sending an empty slice so the API actually clears them instead of retaining orphans. (`IsNewResource()` is no longer needed, a new resource with `secrets_wo` block registers as a change.)

```suggestion
	} else if data.HasChange("secrets_wo") || data.HasChange("secrets_wo_version") {
		// secrets_wo names are tracked in state, so HasChange("secrets_wo") catches
		// adds/renames/removals; secrets_wo_version catches value-only rotations.
		secretsWO := config.GetAttr("secrets_wo")
		switch {
		case !secretsWO.IsNull() && secretsWO.LengthInt() > 0:
			action.Secrets = expandActionSecretsWO(secretsWO)
		case data.HasChange("secrets_wo"):
			// All secrets_wo entries were removed; send an empty slice so the API
			// clears them instead of silently retaining orphaned secrets.
			action.Secrets = &[]management.ActionSecret{}
		}
	}
```

internal/auth0/action/expand.go
```diff
@@ -102,6 +107,20 @@ func expandActionSecrets(secrets cty.Value) *[]management.ActionSecret {
 	return &actionSecrets
 }
 
+func expandActionSecretsWO(secrets cty.Value) *[]management.ActionSecret {
```
bkiran6398: `expandActionSecretsWO` is very identical to existing `expandActionSecrets`. Check if same func can be reused.

internal/auth0/action/expand.go
```diff
@@ -139,7 +158,7 @@ func expandTriggerBindings(config cty.Value) []*management.ActionBinding {
 }
 
 func preventErasingUnmanagedSecrets(ctx context.Context, data *schema.ResourceData, api *management.Management) diag.Diagnostics {
-	if !data.HasChange("secrets") {
+	if !data.HasChange("secrets") && !data.HasChange("secrets_wo_version") {
 		return nil
 	}
```
bkiran6398: The guard must also run on structural `secrets_wo` changes
```suggestion
	if !data.HasChange("secrets") && !data.HasChange("secrets_wo") && !data.HasChange("secrets_wo_version") {
		return nil
	}
```

internal/auth0/action/expand.go
```diff
@@ -148,21 +167,26 @@ func preventErasingUnmanagedSecrets(ctx context.Context, data *schema.ResourceDa
...
+		}
 	}
 
-	if newSecrets != nil {
-		secretsList = append(secretsList, newSecrets.(*schema.Set).List()...)
+	if data.HasChange("secrets_wo_version") {
+		if secretsWO, ok := data.GetOk("secrets_wo"); ok {
+			secretsList = append(secretsList, secretsWO.([]interface{})...)
+		}
 	}
```
bkiran6398: Mirroring the existing `secrets` branch just above. Aggregate **both old and new** names. Including the old names is what lets a user legitimately remove an entry without it being flagged as unmanaged.

```suggestion
	if data.HasChange("secrets_wo") || data.HasChange("secrets_wo_version") {
		// Aggregate both old and new secrets_wo names. Including the old names lets
		// a user remove an entry without it being flagged as an unmanaged secret.
		oldSecretsWO, newSecretsWO := data.GetChange("secrets_wo")
		if oldSecretsWO != nil {
			secretsList = append(secretsList, oldSecretsWO.([]interface{})...)
		}
		if newSecretsWO != nil {
			secretsList = append(secretsList, newSecretsWO.([]interface{})...)
		}
	}
```

internal/auth0/action/expand.go
bkiran6398: Try setting AttributePath conditionally as `secretKeysInConfigMap` can be either set in `secrets` or `secrets_wo`.

bkiran6398 (changes requested):

Asarew:
Thanks for the thorough review @bkiran6398 — pushed in 1a387920. Summary of what changed:

- `expandAction` now triggers on `HasChange("secrets_wo") || HasChange("secrets_wo_version")` and sends `&[]ActionSecret{}` when all entries are removed, so the orphaned-secret scenario you reproduced no longer happens.
- Dropped `expandActionSecretsWO` and reuse `expandActionSecrets` directly.
- `preventErasingUnmanagedSecrets` short-circuit now also considers `secrets_wo`.
- The `secrets_wo` branch in the guard mirrors the `secrets` branch — `GetChange` for both old + new names — so removing an entry no longer hits a false "Unmanaged Action Secret" error.
- Threaded an `attributePath` argument through `checkForUnmanagedActionSecrets` and set it to `"secrets"` or `"secrets_wo"` based on which block triggered. The existing `TestCheckForUntrackedActionSecrets` and the new `TestCheckForUntrackedActionSecretsWithSecretsWO` both cover that.

bkiran6398 (dismissed):

KartikJha (dismissed):
LGTM

KartikJha (commented):
LGTM

KartikJha (approved):
LGTM

Asarew:
Thanks everybody <3

PR #1521, opened 2026-04-06
fix: Add opt-in parameters to eliminate over-fetching in data sources

AkxenTech:
## Problem

Four data sources unconditionally fetch related sub-resources on every read, regardless of whether the caller references them. This causes 2–4x more API calls than necessary and silently consumes rate limit quota.

**Example:** A simple `data \"auth0_user\"` lookup that only needs `email` triggers 3 API calls:
- `GET /api/v2/users/{id}` ✅ needed
- `GET /api/v2/users/{id}/roles` ❌ not requested
- `GET /api/v2/users/{id}/permissions` ❌ not requested

At scale (100 users × 3 calls = 300 requests), this consumes 30–50% of the rate limit window before any actual work is done — the root cause of the "48% problem" reported by large enterprise tenants hitting silent rate limit hangs.

## Solution

Add opt-in boolean parameters (default `false`) to each affected data source. Sub-resources are only fetched when explicitly requested. Existing configurations continue to work — they just no longer make unnecessary API calls.

### Changes per data source

| Data Source | New Parameters | API Calls: Before → After |
|-------------|----------------|---------------------------|
| `auth0_user` | `fetch_roles`, `fetch_permissions` | 3 → 1 (67% reduction) |
| `auth0_role` | `fetch_permissions`, `fetch_users` | 3 → 1 (67% reduction) |
| `auth0_organization` | `fetch_connections`, `fetch_members`, `fetch_client_grants` | 4 → 1 (75% reduction) |
| `auth0_connection` | `fetch_enabled_clients` | 2 → 1 (50% reduction) |

**Overall: ~64% fewer API calls for data source operations.**

### Usage example

```hcl
# Before (implicit, always fetches roles + permissions):
data "auth0_user" "example" {
  user_id = "auth0|abc123"
}

# After (explicit opt-in, 1 API call instead of 3):
data "auth0_user" "example" {
  user_id = "auth0|abc123"
}

# When you actually need roles:
data "auth0_user" "example" {
  user_id     = "auth0|abc123"
  fetch_roles = true
}
```

### Additional changes

- **Rate limit diagnostic logging** (`internal/config`): logs a warning when remaining quota drops below 10% or 5%, and emits a usage summary at the end of the run (total requests, rate limit hits, efficiency score).
- **Rate limit telemetry** (`internal/config`): thread-safe metrics collection for total requests, wait duration, and min/max remaining quota.

---

## Test Results

[...]

Issue #1488, opened 2026-02-27
Auth0 rate limits being hit when applying terraform with a large number of Applications/Clients


TerryAtOctopus:
### Checklist

- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

We have 3500+ Applications/Clients in Auth0, most of which are added dynamically (i.e. not deployed by terraform). We're finding that when we run terraform apply, even when there are no changes to be deployed, we're hitting Auth0 global rate limits (both burst and total limits). The rate limit errors indicate that the APIs causing the problem are to do with getting information about individual clients, so our assumption is that the terraform provider is needing to retrieve each client individually, perhaps in order to determine if there are any changes.

We haven't seen any deployments fail because of this, so the terraform client seems to be handling them correctly. However, when these limits are hit it potentially impacts other functionality.

### Describe the ideal solution

It would be great if there was a way to rate limit on the client side - e.g. to tell the terraform provider that it's not allowed to make more than x requests per minute.

### Alternatives and current workarounds

Do nothing and accept the temporary impact to other Auth0 requests.

MSunyaikin:
We're running into this as well, and I'd argue it's a general problem with the provider rather than something specific to clients/applications — any resource type that exists in larger numbers triggers it.

For us the worst offender is `auth0_prompt_custom_text`. Each combination of prompt and language is a separate resource requiring its own API call. We customize 5 prompts in 21 languages, which already means 105 read requests on every plan/apply before any actual changes are even considered — and this multiplies with each additional prompt or language.

The real impact goes beyond Terraform itself, a deployment exhausts the quota for every consumer of the Management API. Our other services that rely on it get blocked while a Terraform run is in progress, so a routine config change effectively causes a partial outage for unrelated functionality.

duedares-rvj:
Hello @TerryAtOctopus @MSunyaikin   👋 
We added certain optimisation along with [v1.47.0](https://github.com/auth0/terraform-provider-auth0/releases/tag/v1.47.0) to reduce over-fetching of certain resources.

These were part of [1521](https://github.com/auth0/terraform-provider-auth0/pull/1521). 
Feel free to review them and suggest us other conditional placeholders over other data-sources which can bring optimisation.

TerryAtOctopus:
Thanks @duedares-rvj, I'll see if we can make use of those.

[...]

PR #1571, opened 2026-05-08
feat: add opt-in opportunistic pre-fetch cache for clients and client grants

emsearcy:
### 🔧 Changes

Add an opt-in opportunistic pre-fetch cache for `auth0_client` and `auth0_client_grant` resource reads.

Large Auth0 deployments (100+ clients, 100+ grants) incur excessive latency and rate-limit exhaustion (especially non-production tenants!) during `terraform plan` CI runs because the provider fetches each resource individually. When pre-fetch is enabled, any individual resource read:

1. Checks an in-memory cache (by resource type + ID).
2. On a cache miss (while pages remain): fetches the **next page** of resources (50/page), stores all results, advances the page cursor.
3. Re-checks the cache — returns the entry if now found.
4. Falls back to a single-resource fetch if still not found.

This balances latency (never waits for all pages) with opportunistic bulk loading.

**Opt-in:**

```hcl
provider "auth0" {
  prefetch = true  # or AUTH0_PREFETCH=true
}
```

The feature is **off by default** — existing users are unaffected.

**Files changed:**

| File | Change |
|------|--------|
| `internal/prefetch/cache.go` | Thread-safe `sync.RWMutex`-backed cache; page-cursor tracking; per-type fetch mutex (prevents TOCTOU races across parallel goroutines); atomic hit/miss/page counters |
| `internal/prefetch/prefetch.go` | `GetClient` and `GetClientGrant` — opportunistic heuristic + fallback |
| `internal/config/config.go` | `prefetchCache` field, `GetPrefetchCache()` accessor, wired in `ConfigureProvider` |
| `internal/provider/provider.go` | `prefetch` bool schema attribute + `AUTH0_PREFETCH` env var |
| `internal/auth0/client/resource.go` | Use prefetch path in `readClient` when cache non-nil |
| `internal/auth0/client/resource_grant.go` | Use prefetch path in `readClientGrant` when cache non-nil |
| `internal/auth0/client/resource_credentials.go` | Use prefetch path in `readClientCredentials` when cache non-nil |
| `docs/guides/opportunistic_prefetch.md` | Usage guide |
| `internal/prefetch/cache_test.go` | Unit tests |

**Known limitations:**

- **Per-process cache only.** OpenTofu/Terraform spawns multiple provider processes for parallelism. Each process has its own independent in-memory cache.
- **Cache is not invalidated.** Populated once per provider process lifetime — correct for `terraform plan` but pre-fetch should not be used if Auth0 state changes mid-run.
- **Page size is fixed at 50**, matching the SDK default.

**Observability:** At `TF_LOG=DEBUG`, a summary line is emitted when each resource type's page list is exhausted:

```
prefetch: client_grant cache exhausted  pages_fetched=3 cached=114 hits=68 misses=4 hit_rate_pct=94.4
```

### 📚 References

- #1488 — rate limits with large client counts (the core problem this addresses)
- #1521 — community PR with similar motivation for data sources

### 🔬 Testing

Unit tests in `internal/prefetch/cache_test.go` cover concurrent access, TOCTOU safety, pagination exhaustion, and fallback behavior.

Manual timing against a non-production Auth0 tenant with ~270 clients and ~165 grants:

| Run | Wall time |
|-----|-----------|
| Upstream (no prefetch) | 4 min 20 sec |
| `AUTH0_PREFETCH=true` | 49 sec |
| **Speedup** | **~5.3×** |

Cache stats from the prefetch run:
- `auth0_client`: hit_rate=95.8% (264 cached, 6 pages fetched)
- `auth0_client_grant`: hit_rate=100.0% (114 cached, 3 pages fetched)

### 📝 Checklist

### DIFF
--- a/docs/guides/opportunistic_prefetch.md
+++ b/docs/guides/opportunistic_prefetch.md
@@ -1,0 +1,58 @@
+---
+page_title: Opportunistic Pre-fetch
+description: |-
+  Reduce API call count for large Auth0 deployments by enabling opportunistic pre-fetch mode.
+---
+
+# Opportunistic Pre-fetch
+
+Large Auth0 deployments — for example, those with 100+ clients or 100+ client grants — can hit
+excessive latency and rate-limit exhaustion during `terraform plan` runs. By default, the provider
+fetches each `auth0_client` and `auth0_client_grant` resource individually, resulting in one API
+call per resource.
+
+**Opportunistic pre-fetch mode** batches those upstream calls using the Auth0 page-based list
+endpoints, dramatically reducing total API call count.
+
+## How it works
+
+When pre-fetch is enabled, any individual resource read follows this heuristic:
+
+1. Check an in-memory cache (keyed by resource ID).
+2. On a cache miss, if pages remain unfetched, fetch the **next page** (50 resources per page),
+   store all results in the cache, and advance the page cursor.
+3. Re-check the cache — return the resource if it is now present.
+4. Fall back to a direct single-resource API call if the resource was still not found.
+
+This approach balances latency (the provider never waits for all pages before returning) with
+opportunistic bulk loading across a `terraform plan` run.
+
+## Enabling pre-fetch
+
+Add `prefetch = true` to your provider block:
+
+```terraform
+provider "auth0" {
+  domain        = var.auth0_domain
+  client_id     = var.auth0_client_id
+  client_secret = var.auth0_client_secret
+  prefetch      = true
+}
+```
+
+Alternatively, set the environment variable:
+
+```shell
+export AUTH0_PREFETCH=true
+```
+
+Pre-fetch is **opt-in** and defaults to `false`. No changes to individual resource configurations
+are required.
+
+## Scope
+
+The initial release covers `auth0_client` and `auth0_client_grant`. The `auth0_client_credentials`
+resource benefits automatically because it reads client data through the same code path.
+
+Additional resource types (for example, `auth0_resource_server`) may be added in future releases
+without a breaking change.
--- a/internal/prefetch/cache_test.go
+++ b/internal/prefetch/cache_test.go
@@ -1,0 +1,95 @@
+package prefetch
+
+import (
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewCache(t *testing.T) {
+	t.Run("it returns an initialised cache", func(t *testing.T) {
+		c := NewCache()
+		assert.NotNil(t, c)
+	})
+}
+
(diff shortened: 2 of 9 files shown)


internal/prefetch/prefetch.go
```diff
@@ -0,0 +1,175 @@
...
+		if v, ok := cache.getEntry(resourceTypeClient, id); ok {
+			cache.unlockFetch(resourceTypeClient)
+			cache.recordHit(resourceTypeClient)
+			return v.(*management.Client), nil
+		}
+
+		if !cache.isExhausted(resourceTypeClient) {
+			page := cache.nextPage(resourceTypeClient)
+
+			list, err := api.Client.List(ctx,
```
duedares-rvj, line 44: List results may not be equivalent to read results. readClient and readClientCredentials now flatten objects returned from Client.List, not Client.Read. Auth0 documents special field/scope behavior for client fields like `client_authentication_methods`, `client_secret`,`encryption_key`, and `signing keys`; using list responses as read substitutes risks missing fields or producing diffs.

docs/guides/opportunistic_prefetch.md
```diff
@@ -0,0 +1,58 @@
...
+
+1. Check an in-memory cache (keyed by resource ID).
+2. On a cache miss, if pages remain unfetched, fetch the **next page** (50 resources per page),
+   store all results in the cache, and advance the page cursor.
+3. Re-check the cache — return the resource if it is now present.
+4. Fall back to a direct single-resource API call if the resource was still not found.
+
+This approach balances latency (the provider never waits for all pages before returning) with
+opportunistic bulk loading across a `terraform plan` run.
+
```
duedares-rvj, line 29: The cache used seems to be currently process-wide and never invalidated. 
updateClient, updateClientGrant, and updateClientCredentials all call read afterward; if the object was cached before the write, state can be flattened from stale data. Clean call out that “Cache is not invalidated”,  but the provider option affects apply too.

internal/prefetch/cache_test.go
```diff
@@ -0,0 +1,225 @@
+package prefetch
```
duedares-rvj, line 1: These tests are essential, but are mostly cache-unit tests and not provider behaviour tests. They focus on cache state/concurrency and does not validate read-vs-list equivalence, stale-cache-after-update, or integration behaviour.

emsearcy:
> List results may not be equivalent to read results [...] using list responses as read substitutes risks missing fields or producing diffs.

Drat, not sure how I missed that. Since this is fundamental to the design on this pre-fetch, it seems like the whole thing is a no-go on that basis alone ... any further field-aware workarounds needed would likely introduce too much complexity to be worth it.

Thanks for the review (and sorry for my late reply!).

PR #1575, opened 2026-05-12
fix(client): allow dynamic login URI placeholders in initiate_login_uri

### LINKED ISSUE #1574: initiate_login_uri validator rejects dynamic placeholders ({organization.metadata.KEY}, {custom_domain.metadata.KEY}) that the Auth0 API supports
## Provider version

`auth0/auth0` v1.46.0 (still present in latest).

## What happened

The Auth0 Management API supports dynamic placeholders in `initiate_login_uri` for tenants that use Organizations or Multiple Custom Domains:

| Placeholder | Source | Example |
|---|---|---|
| `{custom_domain.metadata.KEY}` | Metadata from the custom domain used in the request | `https://{custom_domain.metadata.public_app_host}/login` |
| `{organization.metadata.KEY}` | Metadata from the Organization associated with the request | `https://{organization.metadata.public_login_host}/login` |

Source: https://auth0.com/docs/authenticate/login/auth0-universal-login/configure-default-login-routes#dynamic-login-uris-with-metadata-placeholders — section "Dynamic login URIs with metadata placeholders". Metadata keys must start with `public_`.

The Terraform provider rejects these values client-side:

```
Error: expected "initiate_login_uri" to be a valid url, got https://{organization.metadata.public_login_host}:
parse "https://{organization.metadata.public_login_host}":
invalid character "{" in host name
```

## Root cause

`internal/auth0/client/resource.go` validates `initiate_login_uri` with `internalValidation.IsURLWithHTTPSorEmptyString`, which calls Go's `net/url.Parse`. Curly braces are not legal in a URL host, so any placeholder of the form `{organization.metadata.KEY}` or `{custom_domain.metadata.KEY}` is rejected before the value ever reaches the API.

## Reproduction

```hcl
resource "auth0_client" "example" {
  name               = "example"
  app_type           = "regular_web"
  initiate_login_uri = "https://{organization.metadata.public_login_host}"
}
```

`terraform validate` (or `terraform plan`) fails with the error above. The same value applied directly to the Auth0 Management API (`PATCH /api/v2/clients/{id}`) is accepted and works correctly at login time.

## Expected behavior

The validator should accept Auth0's documented placeholder syntax. Possible approaches:

1. Skip URL validation when the raw string contains `{organization.metadata.` or `{custom_domain.metadata.`.
2. Substitute placeholders with a stub host before validation (e.g. `{organization.metadata.public_login_host}` → `placeholder`) and parse the result.
3. Replace `IsURLWithHTTPSorEmptyString` with a validator that allows the documented placeholder grammar in the host component.

Option 1 is the least intrusive and matches what other providers do when a field accepts both URLs and templating.

## Workaround

Manage `initiate_login_uri` out-of-band:

```hcl
resource "auth0_client" "example" {
  # ... initiate_login_uri omitted ...
  lifecycle { ignore_changes = [initiate_login_uri] }
}

resource "terraform_data" "example_initiate_login_uri" {
  input = {
    client_id          = auth0_client.example.client_id
    initiate_login_uri = "https://{organization.metadata.public_login_host}"
  }
  provisioner "local-exec" {
    command = "./set-initiate-login-uri.sh ${self.input.client_id} '${self.input.initiate_login_uri}'"
  }
}
```

…where the script obtains a Management API token and `PATCH`es `/api/v2/clients/{id}` directly. Functional, but defeats the purpose of using the provider for this field.

florentinchaussoy:
Closes #1574.

## Why

The Auth0 Management API supports dynamic placeholders in `initiate_login_uri` for tenants that use Organizations or Multiple Custom Domains:

| Placeholder | Source |
|---|---|
| `{organization.metadata.public_*}` | Metadata from the Organization associated with the request |
| `{custom_domain.metadata.public_*}` | Metadata from the custom domain used in the request |

See [Application Settings → Dynamic login URIs with metadata placeholders](https://auth0.com/docs/get-started/applications/application-settings).

The provider validates `initiate_login_uri` with `IsURLWithHTTPSorEmptyString`, which calls Go's `net/url.Parse`. Curly braces are illegal in a URL host, so any value of the form `https://{organization.metadata.public_login_host}` is rejected client-side before reaching the API:

```
Error: expected "initiate_login_uri" to be a valid url, got https://{organization.metadata.public_login_host}:
parse "https://{organization.metadata.public_login_host}":
invalid character "{" in host name
```

Users currently have to manage the field out-of-band (e.g. via a `local-exec` calling the Management API) which defeats the point of using the provider for it.

## What

- Add `IsHTTPSURLOrEmptyStringWithDynamicLoginURIPlaceholders` in `internal/validation`. It substitutes the two documented placeholder forms with a stable host before invoking `net/url.Parse`, then enforces the same HTTPS-only, non-empty-host rules as `IsURLWithHTTPSorEmptyString`.
- Apply the new validator only to the `initiate_login_uri` field on `auth0_client`. The other 5 call sites of `IsURLWithHTTPSorEmptyString` are unchanged — the placeholder syntax is documented only for `initiate_login_uri`, so the more permissive validator stays scoped.
- Update the field description and regenerate docs.

Scoping rationale: the only known callers of these placeholders are the `initiate_login_uri` field, per the docs. Touching the shared validator would silently allow `{...}` in many other URL fields that should remain strict.

## Tests

`internal/validation/validation_test.go` adds `TestIsHTTPSURLOrEmptyStringWithDynamicLoginURIPlaceholders` covering:
- plain HTTPS URLs and empty string still accepted
- `{organization.metadata.public_*}` accepted as host and embedded in path
- `{custom_domain.metadata.public_*}` accepted
- both placeholder types combined in a single URL
- HTTP scheme still rejected (even with a placeholder)
- metadata keys without the `public_` prefix still rejected
- unsupported placeholder sources still rejected
- non-string inputs still rejected

`make test-unit` passes.

## How to test

```hcl
resource "auth0_client" "example" {
  name               = "example"
  app_type           = "regular_web"
  initiate_login_uri = "https://{organization.metadata.public_login_host}/login"
}
```

Before this PR, `terraform validate` fails on the URL grammar. After, it accepts the value and the placeholder is forwarded to the Management API.

## Notes

- No CHANGELOG entry — assuming maintainers add it during the next release cut. Happy to add one if preferred.
- Commits are not GPG-signed; let me know if signing is mandatory for merge and I'll re-sign.


internal/validation/validation.go
```diff
@@ -3,9 +3,17 @@ package validation
...
+	"regexp"
 	"strings"
 )
 
+// dynamicLoginURIPlaceholderRegex matches the Auth0-documented placeholder
+// syntax allowed in initiate_login_uri for tenants using Organizations or
+// Multiple Custom Domains. Metadata keys must start with `public_`.
+// See https://auth0.com/docs/get-started/applications/application-settings
+// (section "Dynamic login URIs with metadata placeholders").
+var dynamicLoginURIPlaceholderRegex = regexp.MustCompile(`\{(organization|custom_domain)\.metadata\.public_[A-Za-z0-9_]+\}`)
```
duedares-rvj: Missing handling for `PUBLIC_`

https://auth0.com/docs/get-started/applications/wildcards-for-subdomains#validation-rules
Public prefix required: The metadata key used in the placeholder must start with public_ or PUBLIC_

ramya18101 (approved):

PR #1585, opened 2026-05-23
fix: field validation for `auth0_self_service_profile_custom_text`

scantisani:
### 🔧 Changes

Validate the `language` and `page` fields for the `auth0_self_service_profile_custom_text` resource to match what the [Set custom text for a self-service profile](https://auth0.com/docs/api/management/v2/self-service-profiles/put-self-service-profile-custom-text) endpoint accepts - currently only `en` and `get-started` respectively.

We got into a mild pickle at $DAYJOB when we tried to add text for a non-`en` language - although the API refused the request, our Terraform state still got updated. The phantom non-`en` resources couldn't be deleted, so we couldn't apply any changes. I ended up modifying the Terraform state manually to get us unstuck.

### 🔬 Testing

I left the existing unit tests for this resource as-is, since it seems like the convention is not to test simple validation functions.

Example of a resource that would now correctly fail validation:

```tf
resource "auth0_self_service_profile_custom_text" "sso_custom_text" {
  sso_id = auth0_self_service_profile.my_self_service_profile.id
  language = "fr" # unsupported language
  page = "unsupported-page" # unsupported page
  body = jsonencode(
    {
	"introduction": "Welcome! This is an updated Text"
    }
  )
}
```

### 📝 Checklist


duedares-rvj (approved):

duedares-rvj:
@scantisani Agghhh, that can certainly be frustrating. But many thanks for raising this PR. 

Ref: https://auth0.com/docs/api/management/v2/self-service-profiles/put-self-service-profile-custom-text
Screenshot 2026-05-29 at 12 00 38 AM

PR #1598, opened 2026-06-02
fix: Handle empty `id_token_signed_response_algs` as unset

bkiran6398:
### 🔧 Changes

**Problem**

The Auth0 API treats this field as three-state: omitting it means "allow all" (default algorithm applied at runtime), a populated list restricts to those algorithms, and an explicit empty array (`[]`) means "reject all" — breaking logins with `unexpected JWT "alg" header parameter`. Customers hit this after auto-generating Terraform config, which exported `id_token_signed_response_algs = []` and sent it verbatim on the next apply.

**Fix**

Omit the field from the API request when the configured list is empty, treating empty the same as unset. The schema and docs description is also updated to note that an unset or empty value applies the default algorithm at runtime.

**Design decision**

Empty and null can't be distinguished in this resource: it's built on `terraform-plugin-sdk/v2`, whose state model collapses null and empty lists to the same representation (`#=0`). The Dashboard avoids the issue by omitting the property from its JSON when nothing is selected rather than sending `[]`. Since an explicit `[]` has no valid use, collapsing it to unset is safe and matches the Dashboard.

### 📚 References

- API Docs: https://auth0.com/docs/api/management/v2/connections/post-connections#body-options-id-token-signed-response-algs-one-of-0

### 🔬 Testing

- Extended the existing `TestAccConnectionOIDC` and `TestAccConnectionOkta` acceptance tests with a step that sets `id_token_signed_response_algs = []` and asserts the apply succeeds with `id_token_signed_response_algs.# = 0`, confirming the empty array is no longer rejected by the API. HTTP recordings were regenerated against a live tenant.
- Manually verified end-to-end: generated config for a connection with the field unset, applied a config containing `id_token_signed_response_algs = []`, and confirmed empty list is not populated in API request.

### 📝 Checklist


ankita10119 (approved):

internal/auth0/connection/expand.go
```diff
@@ -1100,6 +1100,15 @@ func expandConnectionOptionsOkta(data *schema.ResourceData, config cty.Value) (i
 	return options, diag.FromErr(err)
 }
 
+// nonEmptyStrings returns nil for a null OR empty list.
+func nonEmptyStrings(v cty.Value) *[]string {
```
ramya18101, line 1104: Future suggestion: Isn't this function being used in couple of places, See if we can put it utils.go and make re-use of it!

PR #1609, opened 2026-06-15
feat: Add invitation_landing_client_id to My Organization API in client

bkiran6398:
### 🔧 Changes

Adds support for the `invitation_landing_client_id` field on the `auth0_client` resource's `my_organization_configuration` block. `invitation_landing_client_id` is the client ID used as the invitation landing page when creating invitations through the My Organization API.

```hcl
resource "auth0_client" "invitation_landing" {
	name               = "Acceptance Test - MyOrgConfig Landing"
	organization_usage = "allow"
}

  name = "My Org App"

  my_organization_configuration {
    # ... other params ...
    invitation_landing_client_id = auth0_client.invitation_landing.client_id    # new param
  }
}
```

The field is optional and inherited by the `auth0_client` / `auth0_clients` data sources.

> **Known limitation — misleading error on invalid `invitation_landing_client_id` during update:**
> If the referenced client ID does not exist, the Management API returns `404 "The Client ID for Invitations could not be found."` On updates, the provider's shared `HandleAPIError` logic treats any `404` as an out-of-band deletion, clears the resource ID from state, and reports success. Terraform then raises the misleading `Error: Provider produced inconsistent result after apply` instead of the API error. This stems from existing shared 404 handling, which cannot distinguish payload-level from resource-level 404s, is not specific to this field, and is **not addressed in this PR**.

### 📚 References

- go-auth0 v1: [adds `InvitationLandingClientID` to `management.MyOrganizationConfiguration`](https://github.com/auth0/go-auth0/pull/793)

### 🔬 Testing

- Extends the existing `TestAccClientMyOrganizationConfiguration` acceptance test to set `invitation_landing_client_id` (referencing a second `auth0_client` with `organization_usage = "allow"`) and assert it round-trips in both the create and update steps.
- HTTP cassette re-recorded against a tenant with both `enable_my_organization_api` and `my_org_member_management` enabled.
- Behavior validated end-to-end with a live tenant (CRUD and import flow).

To run locally:
```bash
make test-acc FILTER=TestAccClientMyOrganizationConfiguration
```

### 📝 Checklist


duedares-rvj (dismissed):

KartikJha (approved):
LGTM

Issue #1622, opened 2026-06-25
`auth0_client_grant`: `terraform plan -generate-config-out` generates invalid config when `allow_all_scopes = true`


alexmsenger:
### Checklist

- [x] I have looked into the [README](https://github.com/auth0/terraform-provider-auth0#readme) and have not found a suitable solution or answer.
- [x] I have looked into the [documentation](https://registry.terraform.io/providers/auth0/auth0/latest/docs) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/terraform-provider-auth0/issues) and have not found a suitable solution or answer.
- [x] I have upgraded to the [latest version](https://github.com/auth0/terraform-provider-auth0/releases/latest) of this provider and the issue still persists.
- [x] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

When running `auth0 tf generate` followed by `terraform plan -generate-config-out=<file>` for a client grant that has `allow_all_scopes: true` on the Auth0 side, the provider generates a `.tf` file containing **both** `allow_all_scopes = true` and `scopes = []`. This combination is then immediately rejected by the provider's own `CustomizeDiff` validator, making it impossible to import such resources via config generation.

The validator in `resource_grant.go` checks `!scopes.IsNull()` to detect whether `scopes` was provided. However, in Terraform's cty type system, an **explicit empty list `[]` is not null** — it is a known, non-null value. So the condition `allowAllScopes && !scopes.IsNull()` fires even for `scopes = []`, which is the value the flatten/read path writes unconditionally from the Auth0 API response (`scope: []`).

There are two contributing issues:

1. **`validateClientGrant`** (`internal/auth0/client/resource_grant.go`): uses `!scopes.IsNull()` but should also check `scopes.LengthInt() > 0` to allow an empty list alongside `allow_all_scopes = true`.
2. **`flattenClientGrant`** (`internal/auth0/client/flatten.go`): unconditionally writes `scopes` from `clientGrant.GetScope()` even when `allow_all_scopes` is `true`, causing the generated config to contain the redundant `scopes = []`.

### Expectation

[...]

PR #1623, opened 2026-06-25
fix: allow empty scopes list when allow_all_scopes is true in auth0_client_grant

alexmsenger:
### 🔧 Changes

Part of #1622

`validateClientGrant` used `!scopes.IsNull()` to detect whether `scopes` was provided, but in Terraform's cty type system an **explicit empty list `[]` is not null** — it is a known, non-null value. This caused `terraform plan -generate-config-out` to generate an immediately invalid configuration for any client grant with `allow_all_scopes = true`.

Root cause: `flattenClientGrant` unconditionally wrote `scopes = []` (returned by the Auth0 API when `allow_all_scopes` is true) into state. That value was then serialised alongside `allow_all_scopes = true` in the generated `.tf` file, which the validator correctly rejects — but shouldn't, for an empty list.

Two changes:

- **`resource_grant.go`**: Extract a `clientGrantScopesConflictWithAllowAll` helper and tighten the conflict guard from `!scopes.IsNull()` to `!scopes.IsNull() && scopes.LengthInt() > 0`. An explicit `scopes = []` is now treated as equivalent to "not provided".
- **`flatten.go`**: Skip writing `scopes` into state when `allow_all_scopes` is `true`, so the generated config never contains the redundant `scopes = []` alongside `allow_all_scopes = true` in the first place.

### 📚 References

Reproducer: run `auth0 tf generate` followed by `terraform plan -generate-config-out=auth0_generated.tf` on a tenant that has a client grant with `allow_all_scopes = true`. The generated file contains both `allow_all_scopes = true` and `scopes = []`, which immediately fails on the next `terraform plan` with:

```
│ Error: `scopes` cannot be provided when `allow_all_scopes` is set to `true`
```

### 🔬 Testing

New unit tests in `internal/auth0/client/flatten_test.go` (run with `make test-unit`):

- `TestClientGrantScopesConflictWithAllowAll` — 5 table-driven cases covering the full truth table of the extracted predicate, including the previously-failing `allow_all_scopes=true + scopes=[]` case.
- `TestFlattenClientGrant_ScopesOmittedWhenAllowAllScopes` — asserts that `flattenClientGrant` leaves `scopes` unset in state when `allow_all_scopes` is `true`.
- `TestFlattenClientGrant_ScopesSetWhenAllowAllScopesFalse` — asserts normal scope population is unaffected.

### 📝 Checklist
- [x] I have added documentation for all new/changed functionality (N/A — no schema changes)


duedares-rvj (dismissed):

ramya18101 (dismissed):

duedares-rvj:
@alexmsenger 👋 
Thank you for your contribution. 
Kindly run `make lint` and ensure lint is addressed. 

Everything else looks good. :)

alexmsenger:
Hi @duedares-rvj & @ramya18101 👋 

Thank you for the review. I forgot about the listing, sorry. I will check the other PRs as well. Changes here should be pushed now. Please let me know if anything else is missing.

duedares-rvj (approved):

Issue #1415, opened 2025-11-18
Add support for write-only attributes in the `auth0_connection` resource


alethenorio:
### Checklist

- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

It would be great if the `auth0_connection` resource would support [write-only](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/write-only) attributes for data which can be considered sensitive to avoid storing it in the terraform state.

### Describe the ideal solution

There are a few different attributes which immediately comes to mind which might make sense to support

* `client_secret`
* `signing_key.key`
* `decryption_key.key`

There are probably more but these are the ones I am aware of.

Here's what a solution would look like using the example of Google Apps found in the current documentation

```hcl
resource "auth0_connection" "google_apps" {
  name                 = "connection-google-apps"
  is_domain_connection = false
  strategy             = "google-apps"
  show_as_button       = false
  options {
    client_id        = ""
    # The client_secret_wo is an write-only resource and never stored in the terraform state
    client_secret_wo    = ""
    client_secret_wo_version = "1"
    domain           = "example.com"
    tenant_domain    = "example.com"
    domain_aliases   = ["example.com", "api.example.com"]
    api_enable_users = true
    scopes           = ["ext_profile", "ext_groups"]
    icon_url         = "https://example.com/assets/logo.png"
    upstream_params = jsonencode({
      "screen_name" : {
        "alias" : "login_hint"
      }
    })
    set_user_root_attributes = "on_each_login"
    non_persistent_attrs     = ["ethnicity", "gender"]
  }
}
```

I guess the tricky part here might be the fact that Auth0 does not store a version of secrets that I am aware. Maybe it is possible to compare the user desired config vs what's in the state and make that a good enough solution?

### Alternatives and current workarounds

None that I am aware of using Terraform and not clicking in the UI

PR #1624, opened 2026-06-25
fix: prevent spurious diffs from redacted secrets in connection and log stream resources

### LINKED ISSUE #1620: auth0_connection `options.client_secret` and auth0_log_stream `sink.http_authorization` overwrite state with redacted empty values, causing spurious diffs
### Checklist

- [x] I have looked into the README and have not found a suitable solution or answer.
- [x] I have looked into the documentation and have not found a suitable solution or answer.
- [x] I have searched the issues and have not found a suitable solution or answer.
- [x] I have upgraded to the latest version of this provider and the issue still persists.
- [x] I have searched the Auth0 Community forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the Auth0 Code of Conduct.

### Description

The Auth0 Management API redacts secret values on read (it returns an empty/redacted string rather than the real secret). Two attributes write that redacted value straight into Terraform state during the read/flatten step, instead of falling back to the previously-known value already in state. The result is a **perpetual/spurious diff**: every `terraform plan` proposes to "change" the secret back, and every apply re-sends it.

The same diff is **unavoidable when importing an existing tenant via config-driven import** (`terraform plan -generate-config-out=...`), because import has no prior state to fall back on — so even with a correct fallback, import additionally needs `ignore_changes` (see caveat below).

The provider already handles other redacted secrets correctly by falling back to prior state (Datadog API key, Mixpanel password, DB `configuration`, SAML `metadata_xml`). These two attributes are simply missing that same fallback.

**Affected resources / attributes**

1. `auth0_connection` — social / OAuth2 `options.client_secret` (OAuth2, Google, GitHub, Facebook, Apple, LinkedIn, etc.)
2. `auth0_log_stream` — HTTP sink `sink.http_authorization`

**Root cause**

The read/flatten functions set the secret directly from the SDK getter with no prior-state fallback.

`internal/auth0/logstream/flatten.go:80-88`:
```go
func flattenLogStreamSinkHTTP(o *management.LogStreamSinkHTTP) interface{} {
	return map[string]interface{}{
		"http_endpoint":       o.GetEndpoint(),
		"http_content_format": o.GetContentFormat(),
		"http_content_type":   o.GetContentType(),
		"http_authorization":  o.GetAuthorization(), // <-- redacted/empty on read, overwrites state
		"http_custom_headers": o.CustomHeaders,
	}
}
```

`internal/auth0/connection/flatten.go` — `client_secret: options.GetClientSecret()` is written directly in 12 social/OAuth2 flatten functions: lines 173 (GitHub), 199 (WindowsLive), 566 (GoogleOAuth2), 593 (GoogleApps), 630 (OAuth2 + Dropbox/BitBucket/Paypal/Twitter/Amazon/Yahoo/Box/Wordpress/Shopify/Line/Custom), 679 (Facebook), 705 (Apple), 733 (Linkedin), 760 (Salesforce + Community/Sandbox), 841 (OIDC), 914 (Okta), 1066 (AzureAD). Schema: `internal/auth0/connection/schema.go:424` (`TypeString, Optional, Sensitive`; no `DiffSuppressFunc`). Example, `flattenConnectionOptionsOAuth2`:
```go
optionsMap := map[string]interface{}{
	"client_id":     options.GetClientID(),
	"client_secret": options.GetClientSecret(), // <-- redacted/empty on read, overwrites state
	...
}
```

**Contrast — the same packages already do this correctly elsewhere** (prefer prior state over the redacted API value):

- `internal/auth0/logstream/flatten.go` — Datadog API key and Mixpanel password, e.g. `"mixpanel_service_account_password": data.Get("sink.0.mixpanel_service_account_password").(string), // Value does not get read back.`
- `internal/auth0/connection/flatten.go` — DB `configuration` and SAML `metadata_xml`, e.g. `"metadata_xml": data.Get("options.0.metadata_xml").(string), // Does not get read back.`

**Suggested fix**

Mirror the existing Datadog / Mixpanel / DB-config / SAML-metadata pattern: prefer the value already in state over the redacted API value.

- `flattenLogStreamSinkHTTP`: it currently has no `*schema.ResourceData`; thread `data` in (the caller `flattenLogStreamSink(data, ...)` already has it) and set `"http_authorization": data.Get("sink.0.http_authorization")`.
- Connection social/OAuth2 flatten funcs: set `"client_secret": data.Get("options.0.client_secret")`. All 12 functions *already* receive a `*schema.ResourceData` argument currently discarded as `_` — the fix is simply to rename `_` → `data` and read prior state. No call-site changes are needed.

**Import caveat**

The prior-state fallback fixes the **steady-state** perpetual diff. It does **not** fix config-driven import on its own, because import builds state from the API response and there is no prior value to fall back to — the secret is still redacted. For imported resources, users will still need `lifecycle { ignore_changes = [options[0].client_secret] }` / `[sink[0].http_authorization]` (or to set the real secret in config) until/unless write-only attributes (#1415) land. Worth documenting alongside the fix.

Related (different symptom/scope): #1291, #1004 (write-side PUT-vs-PATCH wiping), #1467 (`auth0_client_credentials`, inverse: persists secret), #1415 (write-only attrs request).

### Expectation

On read, when the API returns a redacted/empty secret, the provider should keep the value already present in state, so a no-op plan stays a no-op (matching how Datadog/Mixpanel/DB-config/SAML-metadata are already handled). Today the provider instead writes the redacted (empty) value into state, the next plan shows a diff restoring the configured secret, apply re-sends it, and the diff returns forever.

### Reproduction

Steady-state perpetual diff:
1. Create an `auth0_connection` with strategy `oauth2` (or any social) with a real `options.client_secret`, and an `auth0_log_stream` of `type = "http"` with a real `sink.http_authorization`. Apply.
2. `terraform plan` again with no config change → plan shows `client_secret` and `http_authorization` being re-set ("(sensitive value)"). It never converges.

Config-driven import (worse — no prior state):
1. Point config at an existing tenant: `terraform plan -generate-config-out=gen.tf`.
2. The generated config + first plan always diff on `client_secret` / `http_authorization` because the API never returns the real secret and there is no prior state to fall back to.

v1.50.0 (also confirmed present on `main`); go-auth0 v1.43.0

### Terraform version

Any

alexmsenger:
### 🔧 Changes

Auth0's management API returns an empty string for write-only secret fields after a resource has been created (`options.client_secret` on connections, `sink.http_authorization` on HTTP log streams). When the provider reads those empty values back into state it replaces the original secret, causing a non-empty diff on every subsequent `terraform plan`.

**`internal/auth0/connection/flatten.go`**

In all 12 connection flatten functions that handle `client_secret` (strategies: `github`, `windows_live`, `google-oauth2`, `google-apps`, `oauth2`, `facebook`, `apple`, `linkedin`, `salesforce`, `oidc`, `okta`, `ad`), replace:
```go
"client_secret": options.GetClientSecret(),
```
with:
```go
"client_secret": data.Get("options.0.client_secret").(string), // Value does not get read back.
```
Function signatures updated from `_ *schema.ResourceData` to `data *schema.ResourceData` accordingly.

**`internal/auth0/logstream/flatten.go`**

In `flattenLogStreamSinkHTTP`, replace:
```go
"http_authorization": o.GetAuthorization(),
```
with:
```go
"http_authorization": data.Get("sink.0.http_authorization").(string), // Value does not get read back.
```
The function gains a `data *schema.ResourceData` parameter; the call site in `flattenLogStreamSink` is updated.

This mirrors the existing pattern already used for `options.metadata_xml` in `flattenConnectionOptionsSAML`, `sink.datadog_api_key` in `flattenLogStreamSinkDatadog`, and `sink.mixpanel_service_account_password` in `flattenLogStreamSinkMixpanel`.

Note: import-time behaviour (no prior state) is out of scope and tracked separately in #1415.

### 📚 References

Fixes #1620

### 🔬 Testing

Two new unit tests added to `internal/auth0/connection/flatten_test.go`:
- `TestFlattenConnectionOptionsOAuth2_ClientSecretFallback` – verifies that a nil `ClientSecret` in the API response is replaced by the value already in state.
- `TestFlattenConnectionOptionsGitHub_ClientSecretFallback` – same check for the GitHub strategy.

```
$ go test ./internal/auth0/connection/ -run TestFlattenConnection -v
=== RUN   TestFlattenConnectionOptions
--- PASS: TestFlattenConnectionOptions (0.00s)
=== RUN   TestFlattenConnectionOptionsEmail
--- PASS: TestFlattenConnectionOptionsEmail (0.00s)
=== RUN   TestFlattenConnectionOptionsOAuth2_ClientSecretFallback
--- PASS: TestFlattenConnectionOptionsOAuth2_ClientSecretFallback (0.00s)
=== RUN   TestFlattenConnectionOptionsGitHub_ClientSecretFallback
--- PASS: TestFlattenConnectionOptionsGitHub_ClientSecretFallback (0.00s)
PASS
ok      github.com/auth0/terraform-provider-auth0/internal/auth0/connection     0.852s
```

Acceptance tests require a live Auth0 tenant and cannot be run in CI without credentials, but the fix can be manually verified by:
1. Creating an `auth0_connection` or `auth0_log_stream` with a secret field set.
2. Running `terraform plan` after the initial apply — the plan should show no changes.

### 📝 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [Unreleased]
+
+BUG FIXES:
+- `resource/auth0_connection` – Prevent spurious diffs on `options.client_secret` caused by Auth0 API returning a redacted empty value; provider now preserves the value already in state ([#PLACEHOLDER](https://github.com/auth0/terraform-provider-auth0/pull/PLACEHOLDER))
+- `resource/auth0_log_stream` – Prevent spurious diffs on `sink.http_authorization` caused by Auth0 API returning a redacted empty value; provider now preserves the value already in state ([#PLACEHOLDER](https://github.com/auth0/terraform-provider-auth0/pull/PLACEHOLDER))
+
 ## v1.50.0
 
 ENHANCEMENTS:
--- a/internal/auth0/connection/flatten.go
+++ b/internal/auth0/connection/flatten.go
@@ -155,7 +155,7 @@ func flattenConnectionOptions(data *schema.ResourceData, connection *management.
 }
 
 func flattenConnectionOptionsGitHub(
-	_ *schema.ResourceData,
+	data *schema.ResourceData,
 	rawOptions interface{},
 ) (interface{}, diag.Diagnostics) {
 	options, ok := rawOptions.(*management.ConnectionOptionsGitHub)
@@ -170,7 +170,7 @@ func flattenConnectionOptionsGitHub(
 
 	optionsMap := map[string]interface{}{
 		"client_id":                options.GetClientID(),
-		"client_secret":            options.GetClientSecret(),
+		"client_secret":            data.Get("options.0.client_secret").(string), // Value does not get read back.
 		"set_user_root_attributes": options.GetSetUserAttributes(),
 		"non_persistent_attrs":     options.GetNonPersistentAttrs(),
 		"scopes":                   options.Scopes(),
@@ -181,7 +181,7 @@ func flattenConnectionOptionsGitHub(
 }
 
 func flattenConnectionOptionsWindowsLive(
-	_ *schema.ResourceData,
+	data *schema.ResourceData,
 	rawOptions interface{},
 ) (interface{}, diag.Diagnostics) {
 	options, ok := rawOptions.(*management.ConnectionOptionsWindowsLive)
@@ -196,7 +196,7 @@ func flattenConnectionOptionsWindowsLive(
 
 	optionsMap := map[string]interface{}{
 		"client_id":                options.GetClientID(),
-		"client_secret":            options.GetClientSecret(),
+		"client_secret":            data.Get("options.0.client_secret").(string), // Value does not get read back.
 		"scopes":                   options.Scopes(),
 		"set_user_root_attributes": options.GetSetUserAttributes(),
 		"non_persistent_attrs":     options.GetNonPersistentAttrs(),
@@ -548,7 +548,7 @@ func flattenConnectionOptionsAuth0(
 }
 
 func flattenConnectionOptionsGoogleOAuth2(
-	_ *schema.ResourceData,
+	data *schema.ResourceData,
 	rawOptions interface{},
 ) (interface{}, diag.Diagnostics) {
 	options, ok := rawOptions.(*management.ConnectionOptionsGoogleOAuth2)
@@ -563,7 +563,7 @@ func flattenConnectionOptionsGoogleOAuth2(
 
 	optionsMap := map[string]interface{}{
 		"client_id":                options.GetClientID(),
-		"client_secret":            options.GetClientSecret(),
+		"client_secret":            data.Get("options.0.client_secret").(string), // Value does not get read back.
 		"allowed_audiences":        options.GetAllowedAudiences(),
 		"scopes":                   options.Scopes(),
 		"set_user_root_attributes": options.GetSetUserAttributes(),
@@ -575,7 +575,7 @@ func flattenConnectionOptionsGoogleOAuth2(
 }
 
 func flattenConnectionOptionsGoogleApps(
-	_ *schema.ResourceData,
+	data *schema.ResourceData,
 	rawOptions interface{},
 ) (interface{}, diag.Diagnostics) {
 	options, ok := rawOptions.(*management.ConnectionOptionsGoogleApps)
@@ -590,7 +590,7 @@ func flattenConnectionOptionsGoogleApps(
 
 	optionsMap := map[string]interface{}{
 		"client_id":                options.GetClientID(),
-		"client_secret":            options.GetClientSecret(),
+		"client_secret":            data.Get("options.0.client_secret").(string), // Value does not get read back.
 		"domain":                   options.GetDomain(),
 		"tenant_domain":            options.GetTenantDomain(),
 		"api_enable_users":         options.GetEnableUsersAPI(),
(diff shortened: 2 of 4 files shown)


bkiran6398:
Hi @alexmsenger, Thank you for taking the time to work on a fix for the [reported issue](https://github.com/auth0/terraform-provider-auth0/issues/1620).

That said, I believe this issue should **not** normally occur for Terraform provider users. Please refer to [this linked comment](https://github.com/auth0/terraform-provider-auth0/issues/1620#issuecomment-4915434439) for more context.

For now, I think it would be best to wait for the issue reporter's response before making this change. If we do confirm that some Terraform provider users can receive redacted responses, then instead of always using `data.Get(...).(string)`, we could conditionally fall back to it only when `options.Get*()` returns an empty value. That way, users who receive the full (non-redacted) response from the API would remain unaffected.

In the meantime, if you have a chance, could you also check whether you're able to reproduce the redacted response on your end? Otherwise, we can wait for the issue reporter to get back with more details.

bkiran6398:
Hi @alexmsenger, Thanks again for your time and contribution.

As the [issue reporter confirmed](https://github.com/auth0/terraform-provider-auth0/issues/1620#issuecomment-4917482375), the discrepancy was caused by missing scopes/permissions in the token used in the Terraform configuration, rather than a bug in the code. Based on that, I'm going ahead and closing this PR.

If you continue to experience the discrepancy even with the correct authentication and permissions configured, please let us know, and we'll be happy to investigate further.

PR #1630, opened 2026-06-29
fix: Prevent state drift for api_behavior and signup_behavior in connection password authentication

### LINKED ISSUE #1621: auth0_connection: password `api_behavior` / `signup_behavior` enums reject the API's empty string `""`, breaking `terraform validate` and config-driven import
### Checklist

- [x] I have looked into the README and have not found a suitable solution or answer.
- [x] I have looked into the documentation and have not found a suitable solution or answer.
- [x] I have searched the issues and have not found a suitable solution or answer.
- [x] I have upgraded to the latest version of this provider and the issue still persists.
- [x] I have searched the Auth0 Community forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the Auth0 Code of Conduct.

### Description

On `auth0_connection`, the nested block `options.authentication_methods.password` has two enum attributes whose schema only accepts a fixed set of non-empty values:

- `api_behavior` — `StringInSlice(["required","optional"])`, default `"required"`
- `signup_behavior` — `StringInSlice(["allow","block"])`, default `"allow"`

When a database connection's password authentication methods were never explicitly configured, the Auth0 Management API returns the **empty string `""`** for these fields. The provider's flatten function writes that `""` straight into Terraform state. Because `""` is not in the allowed enum set and is not reconciled with the schema default, this produces two problems:

1. **`terraform validate` fails** on config generated by import: `expected ...api_behavior to be one of ["required" "optional"], got ` (and the analogous error for `signup_behavior`).
2. **A perpetual / non-convergent plan diff**: the API's `""` in state vs. the schema default `"required"` / `"allow"`. This is especially visible with config-driven import (`terraform plan -generate-config-out`).

This is the same class of bug as #1594 (`auth0_attack_protection.bot_detection.bot_detection_level` rejecting `""`), but for different attributes on a different resource. That issue was closed with a wording-only change (#1595 edited the docs and a one-line `Description` string in `resource.go`, but did not change the validator or normalize the value). A wording change cannot fix this case, because here the `""` originates from the **API response during import/read**, not from user-authored config.

**Affected resource / attributes**

- Resource: `auth0_connection` (database / `auth0` strategy)
- Block: `options.authentication_methods.password`
- Attributes: `api_behavior`, `signup_behavior`

**Root cause**

Schema — `internal/auth0/connection/schema.go:314` (`api_behavior`) and `:321` (`signup_behavior`), introduced in PR #1494:
```go
"api_behavior": {
    Type:         schema.TypeString,
    Optional:     true,
    Default:      "required",
    ValidateFunc: validation.StringInSlice([]string{"required", "optional"}, false),
},
"signup_behavior": {
    Type:         schema.TypeString,
    Optional:     true,
    Default:      "allow",
    ValidateFunc: validation.StringInSlice([]string{"allow", "block"}, false),
},
```

Flatten — `internal/auth0/connection/flatten.go:356` (`flattenAuthenticationMethodPassword`; offending getters at lines 364/365). The `if passwordAuthenticationMethod == nil` guard at line 357 only short-circuits a fully-absent block — it does nothing for the `""` case:
```go
func flattenAuthenticationMethodPassword(passwordAuthenticationMethod *management.PasswordAuthenticationMethod) interface{} {
    if passwordAuthenticationMethod == nil {
        return nil
    }
    return []map[string]interface{}{
        {
            "enabled":         passwordAuthenticationMethod.GetEnabled(),
            "api_behavior":    passwordAuthenticationMethod.GetAPIBehavior(),    // returns "" when unset by API
            "signup_behavior": passwordAuthenticationMethod.GetSignupBehavior(), // returns "" when unset by API
        },
    }
}
```

`GetAPIBehavior()` / `GetSignupBehavior()` return `""` when the API never set the value (both back a `*string` in go-auth0 v1.43.0). There is no `DiffSuppressFunc` anywhere in the connection schema reconciling `""` with the default, so the empty string is written verbatim and then fails validation and diffs forever.

**Suggested fix**

Recommended: normalize in flatten. In `flattenAuthenticationMethodPassword`, map `""` to the schema default before writing state:
```go
apiBehavior := passwordAuthenticationMethod.GetAPIBehavior()
if apiBehavior == "" {
    apiBehavior = "required"
}
signupBehavior := passwordAuthenticationMethod.GetSignupBehavior()
if signupBehavior == "" {
    signupBehavior = "allow"
}
```
This fixes **both** symptoms (validate failure *and* the perpetual diff), needs no schema change, and matches a convention already used throughout this file for the identical "API returns empty string for an unset field" situation — e.g. `if options.GetSetUserAttributes() == "" { ... }` (lines 607, 1043, 1087, 1123, 1178) and `GetSigningCert() == ""` (line 1251).

Weaker alternatives: (a) add `""` to the `StringInSlice` sets — fixes validation only, leaves the `""`-vs-default diff; (c) add a `DiffSuppressFunc` — heavier and non-idiomatic here, and still needs (a) for validation.

Note: the wording-only approach used for the analogous #1594 (#1595) does not apply here, since the empty string comes from the API on import/read rather than from user config.

### Expectation

Importing a DB connection whose password authentication methods were never explicitly configured should produce config that passes `terraform validate` and yields an empty (convergent) plan. Today it instead fails `terraform validate` (`expected ...api_behavior to be one of ["required" "optional"], got `) and, even after the value is hand-edited to a valid one, shows a perpetual diff between the API's `""` and the schema default.

### Reproduction

1. Have (or create) a database `auth0_connection` whose `options.authentication_methods.password` was never explicitly configured (the API stores `""` for `api_behavior` / `signup_behavior`).
2. Config-driven import:
   ```hcl
   import {
     to = auth0_connection.db
     id = "con_XXXXXXXX"
   }
   ```
   ```
   terraform plan -generate-config-out=generated.tf
   ```
3. `terraform validate` against `generated.tf` → fails with `expected ...api_behavior to be one of ["required" "optional"], got ` (and `signup_behavior`).
4. Even after manually setting valid values, `terraform plan` shows a perpetual diff (`"" -> "required"` / `"" -> "allow"`).

v1.50.0 (also confirmed present on `main`); go-auth0 v1.43.0

### Terraform version

Any (reproduced with config-driven import)

alexmsenger:
### 🔧 Changes

When the Auth0 API omits `api_behavior` or `signup_behavior` from the password authentication method response (relying on server-side defaults), `flattenAuthenticationMethodPassword` was writing an empty string `""` into Terraform state. Since the schema declares `Default: "required"` and `Default: "allow"` respectively, this caused a perpetual plan diff — Terraform would see `""` from state vs. the schema default, producing a non-empty plan on every run without any actual configuration change.
The fix applies the same default substitution pattern used for `auth0_tenant` (#1587): if the API returns an empty string for either field, the corresponding schema default is substituted before writing to state.

### 📚 References

Closes #1621

### 🔬 Testing

`TestFlattenAuthenticationMethodPassword` in `internal/auth0/connection/flatten_test.go `covers: nil input, explicitly-set values echoed unchanged, and each default fallback independently. Run with `make test-unit`.

### 📝 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ ENHANCEMENTS:
 - `resource/auth0_client` – Add support for configuring `invitation_landing_client_id` in the My Organization API ([#1609](https://github.com/auth0/terraform-provider-auth0/pull/1609))
 - `resource/auth0_client_credentials` – Add support for the PKCS#1 (`RSA PUBLIC KEY`) public key format when computing JWK thumbprints ([#1608](https://github.com/auth0/terraform-provider-auth0/pull/1608))
 
+BUG FIXES:
+- `resource/auth0_connection` – Prevent state drift for `api_behavior` and `signup_behavior` in the password authentication method when the Auth0 API omits these fields, returning server-side defaults ([#1621](https://github.com/auth0/terraform-provider-auth0/pull/1621))
+
 ## v1.49.0
 
 ENHANCEMENTS:
--- a/internal/auth0/connection/flatten.go
+++ b/internal/auth0/connection/flatten.go
@@ -358,11 +358,21 @@ func flattenAuthenticationMethodPassword(passwordAuthenticationMethod *managemen
 		return nil
 	}
 
+	apiBehavior := passwordAuthenticationMethod.GetAPIBehavior()
+	if apiBehavior == "" {
+		apiBehavior = "required"
+	}
+
+	signupBehavior := passwordAuthenticationMethod.GetSignupBehavior()
+	if signupBehavior == "" {
+		signupBehavior = "allow"
+	}
+
 	return []map[string]interface{}{
 		{
 			"enabled":         passwordAuthenticationMethod.GetEnabled(),
-			"api_behavior":    passwordAuthenticationMethod.GetAPIBehavior(),
-			"signup_behavior": passwordAuthenticationMethod.GetSignupBehavior(),
+			"api_behavior":    apiBehavior,
+			"signup_behavior": signupBehavior,
 		},
 	}
 }
--- a/internal/auth0/connection/flatten_test.go
+++ b/internal/auth0/connection/flatten_test.go
@@ -50,3 +50,53 @@ func TestFlattenConnectionOptionsEmail(t *testing.T) {
 		t.Errorf("Expected no diagnostic warnings, got %v", diags)
 	}
 }
+
+func TestFlattenAuthenticationMethodPassword(t *testing.T) {
+	t.Run("nil input returns nil", func(t *testing.T) {
+		result := flattenAuthenticationMethodPassword(nil)
+		if result != nil {
+			t.Errorf("Expected nil for nil input, got %v", result)
+		}
+	})
+
+	t.Run("fields explicitly set are echoed back unchanged", func(t *testing.T) {
+		apiBehavior := "optional"
+		signupBehavior := "block"
+		enabled := true
+
+		result := flattenAuthenticationMethodPassword(&management.PasswordAuthenticationMethod{
+			Enabled:        &enabled,
+			APIBehavior:    &apiBehavior,
+			SignupBehavior: &signupBehavior,
+		})
+
+		m := result.([]map[string]interface{})[0]
+		if m["api_behavior"] != "optional" {
+			t.Errorf("Expected api_behavior=optional, got %q", m["api_behavior"])
+		}
+		if m["signup_behavior"] != "block" {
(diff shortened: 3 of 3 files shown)


bkiran6398:
Hi @alexmsenger, Thank you so much for taking the time to submit a fix.

I wasn't able to reproduce the issue locally, even after testing with several tenants. Regardless of the password field values, the API consistently returned the default values for `api_behavior` and `signup_behavior`. I also tried manually unsetting these fields through the Management API to reproduce the behavior, but wasn't able to do so.

That said, it's entirely possible this issue affects some tenants, including yours. And the provided fix looks good to me.

bkiran6398 (approved):

Issue #1573, opened 2026-05-11
auth0_tenant: cannot clear default_redirection_uri via Terraform config


Greybird:
### Checklist

- [x] I have looked into the [README](https://github.com/auth0/terraform-provider-auth0#readme) and have not found a suitable solution or answer.
- [x] I have looked into the [documentation](https://registry.terraform.io/providers/auth0/auth0/latest/docs) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/terraform-provider-auth0/issues) and have not found a suitable solution or answer.
- [x] I have upgraded to the [latest version](https://github.com/auth0/terraform-provider-auth0/releases/latest) of this provider and the issue still persists.
- [x] I have searched the [Auth0 Community](https://community.auth0.com/c/sdks/5) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

It is currently impossible to clear `default_redirection_uri` on an auth0_tenant resource via Terraform. Setting the field to `""` or `null` in config produces a no-op — the remote value is never cleared.
Plan shows no-op, remote value unchanged

**Initial AI analysis is as follow:**
### Root cause
`default_redirection_uri` is declared as `Optional` + `Computed` in the schema ([`internal/auth0/tenant/resource.go`](https://github.com/auth0/terraform-provider-auth0/blob/main/internal/auth0/tenant/resource.go)):

```tf
"default_redirection_uri": {
    Type:         schema.TypeString,
    Optional:     true,
    Computed:     true,
    ValidateFunc: internalValidation.IsURLWithHTTPSorEmptyString,
    Description:  "The default absolute redirection URI. Must be HTTPS or an empty string.",
},
```
With `Computed: true`, the Terraform Plugin SDK treats an empty string or omitted value as "let the provider own this field". No diff is generated, so `expandTenant` is never called with the new value, and the API is never updated.

The field is also absent from `fetchNullableFields` in expand.go, so the existing two-pass null-PATCH mechanism (already used for `acr_values_supported`, `mtls`, `error_page`, `default_token_quota`, and `skip_non_verifiable_callback_uri_confirmation_prompt`) is not applied to this field.

### Suggested fixes

[...]

PR #1634, opened 2026-06-30
fix: redirection url nullable support

KartikJha:
# fix(tenant): make default_redirection_uri nullable and non-computed

### 🔧 Changes

**Breaking change** — `auth0_tenant.default_redirection_uri` is no longer `Computed`.

Previously the field had `Computed: true`, which caused Terraform to silently ignore the field when it was omitted from config, even if the remote tenant had a non-empty value. This meant users could not clear the field by omitting it, and plan output would show no diff when there was a real divergence.

This PR:

- Removes `Computed: true` from the `default_redirection_uri` schema field so setting it to `""` (_empty string_) in config is treated as an intentional clear and generates a visible plan diff.

**Migration note for existing users:** After upgrading, running `terraform plan` may show a diff proposing to clear `default_redirection_uri` if the field is omitted from config but the remote tenant has a value set. To preserve the existing value, add `default_redirection_uri = "<current value>"` to your configuration before applying or if you want to unset the value use `default_redirection_uri = ""`

### 📚 References

- GH Issue: https://github.com/auth0/terraform-provider-auth0/issues/1573

### 🔬 Testing

- New acceptance test `TestAccTenant_DefaultRedirectionURI` covers all three clearing paths:
  1. Set → plan with cleared value → apply clear (`= ""`) → idempotence check
  2. Re-set → plan with field omitted → apply omit (null-config path via `fetchNullableFields`) → idempotence check
- Regression guard added to `TestAccTenant_Main`: a `PlanOnly: true, ExpectNonEmptyPlan: true` step verifies that removing `default_redirection_uri` from config always generates a diff (catches any future re-introduction of `Computed: true`).

### 📝 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,7 @@
 
 BREAKING CHANGES:
 
-* `auth0_tenant`: The `default_redirection_uri` field is no longer `Computed`. Users who
-  omit this field from their configuration but have a non-empty value set on the remote
-  tenant will see a plan diff on the next `terraform plan`. To prevent an unintended clear,
-  explicitly add `default_redirection_uri = "<current value>"` to your configuration before
-  upgrading.
+* `auth0_tenant`: The `default_redirection_uri` field is no longer `Computed`. Users who omit this field from their configuration but have a non-empty value set on the remote tenant will see a plan diff on the next `terraform plan`. To prevent an unintended clear, explicitly add `default_redirection_uri = "<current value>"` to your configuration before upgrading.
 
 ## v1.51.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 ## [Unreleased]
 
+BREAKING CHANGES:
+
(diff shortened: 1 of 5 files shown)


bkiran6398 (changes requested):
The primary goal is to provide users with a way to unset `default_redirection_uri` via terraform provider.

To minimize the risk of breaking changes, could we treat only an explicitly configured empty string (`""`) as a request to unset the value? That way, the existing behavior for `null` or an omitted `default_redirection_uri` remains a no-op, ensuring there is no impact on users who have never configured this attribute.

KartikJha:
> The primary goal is to provide users with a way to unset `default_redirection_uri` via terraform provider.
> 
> To minimize the risk of breaking changes, could we treat only an explicitly configured empty string (`""`) as a request to unset the value? That way, the existing behavior for `null` or an omitted `default_redirection_uri` remains a no-op, ensuring there is no impact on users who have never configured this attribute.

Hey @bkiran6398 ,

I have made the suggested changes please review.

bkiran6398 (approved):

PR #1636, opened 2026-07-01
fix: Rotate private_key_jwt client_credentials without exceeding the max cap

bkiran6398:
### 🔧 Changes

Fixes a `400 Bad Request: A client can have a maximum of N credentials` error. It happened when replacing all `private_key_jwt` credentials on an `auth0_client_credentials` resource in a single apply.

**Root cause:** The update created all new credentials before deleting the old ones. Auth0 enforces the credential limit at creation time. So a full swap briefly exceeded the limit and was rejected.

**Fix: interleaved rotation:** Credentials are now swapped one slot at a time. Each removal is paired with an addition, and the order within a pair adapts to the live attached count. At capacity, a slot is freed (detach + delete) before it is filled (create + attach), so the count never overshoots the limit. With headroom below the minimum cap, the new credential is added before the old one is removed, so a valid credential always stays attached. The limit value is not needed, which matters, since the API does not expose it. Rotation is planned into ordered steps, then applied against a live snapshot of attached credentials.

Rotating `oldA, oldB` to `newA, newB` at capacity (credential limit 2):

```
before:  oldA, oldB
         → add newA, newB   → oldA, oldB, newA, newB   ✗ 4 > limit, rejected
         → remove oldA, oldB

after:   oldA, oldB
         → remove oldB       → oldA
         → add newA          → oldA, newA
         → remove oldA       → newA
         → add newB          → newA, newB   ✓ never exceeds 2
```

Rotating a single credential `old` to `new` (headroom below the cap) adds first, so the client is never left with zero credentials:

```
after:   old
         → add new           → old, new
         → remove old        → new         ✓ never drops to 0
```

> **Continuity is maintained:** at least one valid credential stays attached to the client throughout the swap, so in-flight authentication keeps working with no availability gap. This preserves the zero-downtime rotation guarantee the original create-before-delete ordering was designed for.

Additional robustness:

- A delete failure (other than a `404`, treated as already gone) stops the rotation instead of being deferred.
- If a newly created credential fails to attach, it is rolled back so it does not linger unattached and consume a slot on the next apply.

### 📚 References

- Zero-downtime guidelines for rotating credentials: https://auth0.com/docs/get-started/applications/rotate-credentials#active-credentials
- Default credential limits: https://auth0.com/docs/get-started/applications/configure-private-key-jwt#credential-limits

### 🔬 Testing

- **Unit tests** (`resource_credentials_internal_test.go`): cover the rotation planner. They check remove-first at capacity, add-first with headroom (single-credential rotation never drops to zero), uneven add/remove counts, pure add and pure remove, ID-less removals, and the count trough/peak invariants.
- **Acceptance test** (`TestAccClientCredentialsPrivateKeyJWTFullRotation`): creates two `private_key_jwt` credentials, then replaces both in one apply. Asserts the resource plans an update and ends with the two new credentials. Recorded HTTP interactions are included.
- Manually verified rotation of clients up-to limit.

### 📝 Checklist


developerkunal (approved):
LGTM

PR #1648, opened 2026-07-14
feat: tfp xaa support added (EA Only)

KartikJha:
# Add Cross-App Access (XAA) requesting-app support to auth0_connection

### 🔧 Changes

Adds support for configuring an `auth0_connection` as a requesting application authorization server for Cross-App Access (XAA). This is an Early Access feature that requires the `token_vault_xaa` flag to be enabled on the tenant, and is only supported on `oidc` and `okta` strategy connections.

- **New `cross_app_access_requesting_app` block** (`schema.go`): a `Max: 1` block with a single required `active` (Boolean) attribute. It is `Optional` + `Computed` so the last-known value sticks in state. Also surfaced on the `auth0_connection` data source (inherited from the resource schema).
- **Expand logic** (`expand.go`): new `expandConnectionCrossAppAccessRequestingApp` reads the raw config and only sends the block when it has changed (`data.HasChange`), avoiding no-op writes. As part of this change, `expandConnectionAuthentication` and `expandConnectionConnectedAccounts` were hardened to guard against null / non-list / empty raw config before iterating.
- **Flatten logic** (`flatten.go`): new `flattenConnectionCrossAppAccessRequestingApp` maps the API response back into state.
- **Sticky-removal behavior**: removing the block from configuration is intentionally a no-op — it does not deactivate the purpose on the connection. Callers must set `active = false` explicitly to deactivate. Documented in the schema description, docs, and example.
- **Dependency bump** (`go.mod`): `github.com/auth0/go-auth0` bumped to `v1.44.1-0.20260714184021-5a0a56dc31ca` to pick up the `CrossAppAccessRequestingApp` model.
- **Docs & example**: regenerated `docs/resources/connection.md` and `docs/data-sources/connection.md`; added `examples/resources/auth0_connection/resource_with_oidc_xaa.tf`.

### 📚 References

- [go v1 PR](https://github.com/auth0/go-auth0/pull/824)

### 🔬 Testing

Covered by new acceptance tests (with recorded HTTP fixtures) in `internal/auth0/connection/resource_test.go`:

- `TestAccConnectionCrossAppAccessRequestingApp` — exercises the full lifecycle: create with `active = true`, update to `active = false`, an unrelated update (`display_name`) that preserves the sticky value, block removal confirming the last-known value stays in state, and confirming the data source inherits the field.
- `TestAccConnectionCrossAppAccessRequestingAppNegative` — asserts the API's `operation_not_supported` error passes through cleanly for unsupported (non-oidc/okta) strategies, rather than being masked or pre-validated client-side.

Recorded cassettes added: `TestAccConnectionCrossAppAccessRequestingApp.yaml`, `TestAccConnectionCrossAppAccessRequestingAppNegative.yaml`, and `TestAccConnectionConnectedAccounts.yaml`.

### 📝 Checklist


internal/auth0/connection/schema.go
```diff
@@ -1665,3 +1666,24 @@ var connectedAccountsSchema = &schema.Schema{
...
+var crossAppAccessRequestingAppSchema = &schema.Schema{
+	Type:     schema.TypeList,
+	Optional: true,
+	Computed: true,
+	MaxItems: 1,
+	Description: "Configure the purpose of a connection to be used as a requesting application " +
+		"authorization server for Cross-App Access (XAA). This is an Early Access feature and " +
+		"requires the `token_vault_xaa` flag to be enabled on your tenant. Only supported on " +
+		"`oidc` and `okta` strategy connections. " +
+		"**Note:** Once configured, removing this block from your configuration is a no-op and will " +
```
bkiran6398, line 1683: Is this no-op behavior inline with API behavior? Doesn't API support removing this block by sending null in PATCH?

internal/auth0/connection/schema.go
```diff
@@ -1665,3 +1666,24 @@ var connectedAccountsSchema = &schema.Schema{
...
 	},
 }
+
+var crossAppAccessRequestingAppSchema = &schema.Schema{
+	Type:     schema.TypeList,
+	Optional: true,
+	Computed: true,
+	MaxItems: 1,
+	Description: "Configure the purpose of a connection to be used as a requesting application " +
+		"authorization server for Cross-App Access (XAA). This is an Early Access feature and " +
```
bkiran6398, line 1680: For early access parameters, append `(EA only)` to the end of the description, following the convention used throughout this repository. Apply the same convention to the PR title as well.

internal/auth0/connection/expand.go
```diff
@@ -129,9 +130,14 @@ func expandConnectionAuthentication(data *schema.ResourceData) *management.Authe
 		return nil
 	}
 
+	rawConfig := data.GetRawConfig().GetAttr("authentication")
```
bkiran6398, line 133: Is this change relevant to this feature?

internal/auth0/connection/expand.go
```diff
@@ -144,9 +150,14 @@ func expandConnectionConnectedAccounts(data *schema.ResourceData) *management.Co
 		return nil
 	}
 
+	rawConfig := data.GetRawConfig().GetAttr("connected_accounts")
```
bkiran6398, line 153: Is this change relevant to this feature?

internal/auth0/connection/expand.go
```diff
@@ -155,6 +166,27 @@ func expandConnectionConnectedAccounts(data *schema.ResourceData) *management.Co
 	return &connectedAccounts
 }
 
+func expandConnectionCrossAppAccessRequestingApp(data *schema.ResourceData) *management.CrossAppAccessRequestingApp {
+	if !data.HasChange("cross_app_access_requesting_app") {
+		return nil
+	}
+
+	rawConfig := data.GetRawConfig().GetAttr("cross_app_access_requesting_app")
+	if rawConfig.IsNull() || !rawConfig.Type().IsListType() || rawConfig.LengthInt() == 0 {
```
bkiran6398: It looks like the `!rawConfig.Type().IsListType()` check is redundant, since this parameter is already declared as `TypeList` in the schema.

internal/auth0/connection/schema.go
KartikJha: Screenshot 2026-07-28 at 10 34 40 AM
API throws 400 on null payload

internal/auth0/connection/expand.go
KartikJha: This is a sanity check strips the null bad parameter early in the flow, for `connection { connected_accounts }` preventing the API 400 response. Since this was a identification in the XAA field support similar expand impls have been updated. 😄 

Refer the earlier comment - https://github.com/auth0/terraform-provider-auth0/pull/1648#discussion_r3662916674

internal/auth0/connection/expand.go
KartikJha: This sanity check strips the null bad parameter early in the flow for `connection { authentication }`, preventing the API 400 response. Since this was a identification in the XAA field support similar expand impls have been updated. 😄 


Refer the earlier comment - https://github.com/auth0/terraform-provider-auth0/pull/1648#discussion_r3662916674

internal/auth0/connection/expand.go
KartikJha: Updated

internal/auth0/connection/schema.go
KartikJha: Updated

internal/auth0/connection/expand.go
bkiran6398: Please remove `!rawConfig.Type().IsListType()` check. Brief the change rationale in PR description.

bkiran6398: And if these changes update the behavior of removing block from `400` to `no-op`, please highlight that in the description like its there for `cross_app_access_requesting_app`.

- Do the same for `authentication` if applicable.

KartikJha: Updated

bkiran6398 (approved):

PR #1653, opened 2026-07-20
feat: Add support to manage `auth0_managed` for Network ACLs

KartikJha:
### 🔧 Changes

Adds support for Auth0-curated blocklists to the `auth0_network_acl` resource and data source via a new `auth0_managed` attribute on rule `match` and `not_match` blocks. This lets a rule match against Auth0-maintained lists (for example `auth0.icloud_relay_proxy` or `auth0.low_reputation`) without the practitioner enumerating IPs or ASNs.

This is an Early Access feature: it requires the `advanced-breached-password-detection` entitlement and the `tenant_acl_curated_blocklists` feature flag to be enabled on the tenant.

- **Schema** (`resource.go`): new `auth0_managed` `TypeList` of strings on the rule match schema, with a `ValidateDiagFunc` requiring the `auth0.<name>` identifier form (e.g. `auth0.icloud_relay_proxy`).
- **Expand** (`expand.go`): reads `auth0_managed` into `NetworkACLRuleMatch.Auth0Managed`, plus a new `validateAuth0ManagedMutualExclusivity` check that errors if `auth0_managed` is set on both `match` and `not_match` in the same rule, mirroring the API invariant.
- **Flatten** (`flatten.go`): surfaces `auth0_managed` back into state on read.
- **Error enrichment** (`resource.go`): create/update wrap 403 responses with a hint naming both required gates (entitlement + feature flag), but only when the rule actually uses `auth0_managed`.
- **Dependency**: bumps `github.com/auth0/go-auth0` to a pre-release build exposing the `Auth0Managed` field on `NetworkACLRuleMatch`.
- **Docs & examples**: documents the new attribute and adds two example resources (a `match` blocklist and a `not_match` allow-unless case demonstrating the mutual-exclusivity boundary).

### 📚 References

go-auth0-v1 [PR](https://github.com/auth0/go-auth0/pull/830)

### 🔬 Testing

- New acceptance tests: `TestAccNetworkACLAuth0Managed`, `TestAccNetworkACLAuth0ManagedValidation` (resource) and `TestAccNetworkACLDataSourceAuth0Managed` (data source), with recorded HTTP fixtures under `test/data/recordings/`.
- The validation test exercises the mutual-exclusivity error path (`auth0_managed` set on both `match` and `not_match`).
- Run with: `make test-acc TESTARGS="-run TestAccNetworkACLAuth0Managed"`.

### 📝 Checklist


internal/auth0/networkacl/resource.go
```diff
@@ -95,6 +99,21 @@ var networkACLRuleMatchSchema = &schema.Schema{
...
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+					ValidateDiagFunc: validation.ToDiagFunc(validation.StringMatch(
+						regexp.MustCompile(`^auth0\.[^\.\s]+$`),
+						"must be an Auth0-curated blocklist identifier of the form 'auth0.<name>' (for example, 'auth0.icloud_relay_proxy')",
+					)),
+				},
+				Description: "Auth0-curated blocklists to match against (for example, `auth0.icloud_relay_proxy`). " +
```
bkiran6398: Add EA only to description and PR title as per the project practice

internal/auth0/networkacl/expand.go
```diff
@@ -91,16 +91,38 @@ func expandNetworkACL(data *schema.ResourceData) (*management.NetworkACL, error)
...
 	}
 
+	if err := validateAuth0ManagedMutualExclusivity(networkACL.Rule.Match, networkACL.Rule.NotMatch); err != nil {
+		return nil, err
+	}
+
 	return networkACL, nil
 }
 
+// validateAuth0ManagedMutualExclusivity ensures the auth0_managed curated blocklists
```
bkiran6398: I'd remove `validateAuth0ManagedMutualExclusivity` and leave this validation to the API.

* The constraint is publicly undocumented. I couldn't find any mention of this invariant in the [Management API docs](https://auth0.com/docs/api/management/v2/network-acls/post-network-acls#body-rule-match-auth0-managed), v2 reference, or OpenAPI spec.
* Similar restrictions for other matcher fields are enforced by the backend rather than the provider. Keeping this validation server-side avoids duplicating API behavior in the provider and reduces maintenance if these validation rules change in the future.

internal/auth0/networkacl/resource.go
```diff
@@ -205,6 +224,42 @@ var networkACLRuleActionSchema = &schema.Schema{
 	},
 }
 
+// enrichAuth0ManagedError augments a 403 from the Management API with a hint
+// naming both gates required for the Early Access auth0_managed curated
+// blocklists, when the rule being sent actually uses that field.
+func enrichAuth0ManagedError(networkACL *management.NetworkACL, err error) error {
```
bkiran6398: `enrichAuth0ManagedError` treats any `403` with `auth0_managed` in the request as an Early Access failure. Could this misclassify authorization failures (e.g. missing scopes) for tenants that already have Early Access enabled? Is there a more specific signal in the response (error code, message, etc.) that distinguishes the EA case from generic `403` auth/scope errors?

KartikJha: Updated

internal/auth0/networkacl/expand.go
KartikJha: Updated

internal/auth0/networkacl/resource.go
KartikJha: Updated

bkiran6398 (approved):

PR #1658, opened 2026-07-24
feat: add Cross App Access support to connection and client (EA only)

bkiran6398:
### 🔧 Changes

Adds Cross-App Access (XAA) resource-application support across the client and connection resources. All fields are Early Access only.

- **`auth0_client`**: new `identity_assertion_authorization_grant` block that opts a client into the Identity Assertion Authorization Grant (ID-JAG) exchange used by XAA. Its required `active` sub-field toggles whether the client can exchange ID-JAGs for access tokens.
- **`auth0_connection`**: new `cross_app_access_resource_app` block letting a connection act as an XAA resource application, with a required `status` of `enabled` or `disabled`.
- **`auth0_connection`**: new `oidc_metadata` option (`oidc`, `okta`, `samlp` strategies) exposing the OIDC discovery document, which the provider never surfaced before.

Both blocks are Optional-only (not Computed) so that removing a block surfaces a diff; because the API drops `omitempty` fields, removal is applied via a raw `PATCH … null` before the typed update. `oidc_metadata` is Computed to absorb the server-populated discovery value, and uses `DiffSuppressFunc` to normalize JSON so the API-enriched superset does not produce perpetual drift.

New fields are also exposed on the corresponding `auth0_client` / `auth0_connection` data sources.

**Usage:**
```hcl
resource "auth0_client" "xaa_app" {
  name = "xaa-app"

  identity_assertion_authorization_grant {
    active = true
  }
}

resource "auth0_connection" "okta" {
  name     = "okta-xaa"
  strategy = "okta"

  cross_app_access_resource_app {
    status = "enabled"
  }
}
```

### 📚 References

- SDK v1: https://github.com/auth0/go-auth0/pull/839

### 🔬 Testing

- Acceptance tests cover create, toggle `active`/`status`, and true block removal for both new blocks, plus `oidc_metadata` create/read, each with recorded HTTP cassettes.
- Unit tests cover the flatten paths for both connection fields and the client block.
- Manually verified that removing a block produces a clean diff and that `oidc_metadata` no longer shows phantom drift on `oidc`/`okta` connections across repeated plans.

Run locally:
```bash
make test-acc FILTER="TestAccClientIdentityAssertionAuthorizationGrant|TestAccConnectionCrossAppAccessResourceApp|TestAccConnectionOIDCMetadata"
```

### 📝 Checklist

### DIFF
--- a/internal/auth0/connection/expand.go
+++ b/internal/auth0/connection/expand.go
@@ -79,6 +79,7 @@ func expandConnection(
 		Authentication:              expandConnectionAuthentication(data),
 		ConnectedAccounts:           expandConnectionConnectedAccounts(data),
 		CrossAppAccessRequestingApp: expandConnectionCrossAppAccessRequestingApp(data),
+		CrossAppAccessResourceApp:   expandConnectionCrossAppAccessResourceApp(data),
 	}
 
 	strategy := data.Get("strategy").(string)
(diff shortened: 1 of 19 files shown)


internal/auth0/connection/expand.go
```diff
@@ -1176,6 +1216,7 @@ func expandConnectionOptionsSAML(_ *schema.ResourceData, config cty.Value) (inte
 		GlobalTokenRevocationJWTSub: value.String(config.GetAttr("global_token_revocation_jwt_sub")),
 		DestinationURL:              value.String(config.GetAttr("destination_url")),
 		RecipientURL:                value.String(config.GetAttr("recipient_url")),
+		DiscoveryURL:                value.String(config.GetAttr("discovery_url")),
```
duedares-rvj, line 1209: So were we missing `DiscoveryURL`  for SAML connection?

duedares-rvj (dismissed):

internal/auth0/connection/expand.go
bkiran6398: No, `DiscoveryURL` support for SAML is added as a new requirement.

KartikJha (approved):
LGTM

PR #1666, opened 2026-08-04
fix: add name to modify creds apply diff check

KartikJha:
# Fix client credentials diff matching to compare cred name

### 🔧 Changes

- `classifyCredentialChanges` in `internal/auth0/client/resource_credentials.go` now requires `name` to match (in addition to PEM/key-id and algorithm) before treating an added credential and a removed credential as the same underlying credential.
- Previously, two `private_key_jwt` credentials with identical PEM/algorithm but different `name` values were matched to each other, so a `name`-only change went undetected and was not applied against the Auth0 API.

### 🔬 Testing

- No new automated tests were added in this change; existing coverage for `classifyCredentialChanges` should be extended to cover the name-mismatch case.
- Manual verification: change only the `name` of an existing `private_key_jwt` credential in configuration and confirm `terraform plan`/`apply` now applies the name update instead of treating it as unchanged.

### 📝 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ BUG FIXES:
 - `resource/auth0_connection` – Fix updates failing with `cannot patch unique property on email attribute` on `auth0` connections whose `options.attributes.email.unique` is `false`. The Management API reads a missing `unique` as `true`, so any change to an unrelated option was rejected ([#1669](https://github.com/auth0/terraform-provider-auth0/pull/1669))
 
 
+BUG FIXES:
+- `resource/auth0_client_credentials` – Include `name` when matching added/removed `private_key_jwt` credentials during diff classification, so credentials that differ only by name are no longer incorrectly treated as the same credential
+
 ## v1.54.0
 
 FEATURES:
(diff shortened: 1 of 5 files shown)


bkiran6398:
* Renaming a single client credential results in the following error:

  ```
  Error: 400 Bad Request: credentials contains public keys that already exist in client: {"name":"Renamed Cred","kid":"w0k...Wu0"}
  ```

* This happens because when a client has only a single credential, `planCredentialRotation` performs the rotation in the order **attach new** → **remove old**. If the provider sends the same public key with only the `name` changed, the API attempts to create the new credential before deleting the existing one. Since both credentials have the same `kid`, the API rejects the request because the public key already exists.

* To reproduce the issue:

  ```hcl
  resource "auth0_client_credentials" "probe" {
    client_id             = auth0_client.probe.id
    authentication_method = "private_key_jwt"

    private_key_jwt {
      credentials {
        name            = "Original Name" # Change to "Renamed Cred", then apply.
        credential_type = "public_key"
        algorithm       = "RS256"
        pem             = file("cert.pem")
      }
    }
  }
  ```

* The same limitation already exists when updating the `algorithm` field, as it follows the same rotation flow. Ideally, the same patch logic should address both cases.

bkiran6398 (dismissed):

bkiran6398 (approved):

PR #1690, opened 2026-08-27
feat: add write-only options_client_secret_wo to connection

bkiran6398:
### 🔧 Changes

Adds a write-only client secret for connection strategies, keeping secrets out of Terraform state/plan and supporting ephemeral sources (secrets manager, Vault, another resource's write-only output).

- `options_client_secret_wo` and `options_client_secret_wo_version` are top-level siblings of `options`, since the SDK forbids write-only fields inside the Computed block.
- Resent on every apply because PATCH replaces options wholesale.
- Flatten blanks the API-echoed secret on the write-only path, preventing it from appearing in `options.client_secret`.

```hcl
resource "auth0_connection" "my_connection" {
  /* ... */ 

  options_client_secret_wo         = var.connection_client_secret # ephemeral
  options_client_secret_wo_version = 1

  options { client_id = "1234567" /* ... */ }
}
```

### 📚 References

- https://github.com/auth0/terraform-provider-auth0/discussions/1450#discussioncomment-18095293
- [Known terraform plugin SDK caveat](https://github.com/hashicorp/terraform-plugin-sdk/issues/1612): `plan -refresh=false` can show a spurious non-empty plan; a regular plan/apply works as expeted. Documented in the resource.

### 🔬 Testing

- Unit tests cover expand/flatten: the write-only path blanks the echoed secret and persists the version; the legacy path is unchanged.
- Acceptance test asserts the plan-time `ConflictsWith` / `RequiredWith` constraints. An apply-based test is intentionally omitted because the harness's post-apply non-refresh plan trips on [terraform-plugin-sdk#1612](https://github.com/hashicorp/terraform-plugin-sdk/issues/1612), which does not affect real usage.
- Verified end-to-end against a live tenant, including rotation and no-leak (secret absent from state).

### 📝 Checklist

### DIFF
--- a/docs/resources/connection.md
+++ b/docs/resources/connection.md
@@ -13,6 +13,12 @@ creation of multiple connections per strategy, the additional connections may no
 
 ~> When updating the `options` parameter, ensure that all nested fields within the `options` schema are explicitly defined. Failing to do so may result in the loss of existing configurations.
 
+~> When `options_client_secret_wo` (write-only) is set, `terraform plan -refresh=false` may report a
+non-empty plan for unrelated optional `options` fields (e.g. `+ scripts = {}`). This is an upstream
+limitation of the Terraform Plugin SDK ([hashicorp/terraform-plugin-sdk#1612](https://github.com/hashicorp/terraform-plugin-sdk/issues/1612))
+that only surfaces without a refresh; a normal `terraform plan`/`apply` (which refreshes) is
+unaffected and idempotent.
+
 ## Example Usage
 
 ### Auth0 Connection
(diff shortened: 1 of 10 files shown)


docs/resources/connection.md
```diff
@@ -13,6 +13,12 @@ creation of multiple connections per strategy, the additional connections may no
 
 ~> When updating the `options` parameter, ensure that all nested fields within the `options` schema are explicitly defined. Failing to do so may result in the loss of existing configurations.
 
+~> When `options_client_secret_wo` (write-only) is set, `terraform plan -refresh=false` may report a
+non-empty plan for unrelated optional `options` fields (e.g. `+ scripts = {}`). This is an upstream
+limitation of the Terraform Plugin SDK ([hashicorp/terraform-plugin-sdk#1612](https://github.com/hashicorp/terraform-plugin-sdk/issues/1612))
```
duedares-rvj, line 18: Does this affect us when we are on SDKv2?

duedares-rvj (approved):

PR #1698, opened 2026-09-04
feat(network_acl, network_acl_keys): hmac signature signal support for tenant acl

KartikJha:
### 🔧 Changes

Introduces HMAC-SHA256 signing key management for Network ACL HTTP Message Signature verification, and extends `auth0_network_acl` rules to match/not-match on those signatures.

**New resource: `auth0_network_acl_key`**
- CRUD lifecycle for `/api/v2/keys/network-acls/*` (EA). Supports `name`, `alg` (`hmac-sha256` only), and `value` (write-only Base64 key material, 32–512 decoded bytes).
- `value` is never stored in Terraform state. Change detection is fingerprint-based: `CustomizeDiff` computes SHA-256 of the decoded key and calls `d.ForceNew("value")` when the fingerprint changes, so key rotation always triggers replacement.
- `fingerprint` (Computed) is stored in state and used for out-of-band drift detection in `Read`.
- Delete surfaces 409 verbatim (key still referenced by an ACL rule) so the user gets the blocking ACL IDs.
- Tenant cap: 10 keys max — docs warn against `create_before_destroy`.

**New data source: `auth0_network_acl_key`**
- Read-only lookup by `id`; returns all non-sensitive fields (`name`, `alg`, `fingerprint`, `created_at`, `updated_at`). No `value`.

**Extended `auth0_network_acl` resource and data sources**
- New optional `http_message_signature` block inside `match` / `not_match`, containing a `keys` list (1–10 items) of `{ id }` references to `auth0_network_acl_key` resources.
- Expand and flatten paths updated; the same shared `flattenNetworkACLRule` function means data source reads pick up the new block automatically.

```hcl
resource "auth0_network_acl_key" "hmac" {
  name  = "my-hmac-key"
  alg   = "hmac-sha256"
  value = var.hmac_key_base64
}

resource "auth0_network_acl" "signed_only" {
  description = "Block unsigned requests"
  active      = true
  priority    = 99
  rule {
    action { block = true }
    scope  = "tenant"
    not_match {
      http_message_signature {
        keys { id = auth0_network_acl_key.hmac.id }
      }
    }
  }
}
```

### 📚 References

- Backed by go-auth0 v3 SDK (`v3.4.0`) — `config.GetAPIV3().Keys.NetworkACLs.{Create,Get,Delete}`.
- `http_message_signature` types (`NetworkACLHTTPMessageSignature`, `NetworkACLHTTPMessageSignatureKey`) are from the pinned go-auth0 v1 commit (`v1.48.1-0.20260901...`) - https://github.com/auth0/go-auth0/pull/857

### 🔬 Testing

- Acceptance tests run and recorded against live tenant `kartik-esd-57610.sus.auth0.com` (feature-flagged for `tenant_acl_hmac_signature`). Cassettes committed to `test/data/recordings/`.
- Tests use priority 99 to avoid collisions with pre-existing ACLs on the tenant.

```bash
# With cassette playback (CI):
make test-acc FILTER="TestAccNetworkACLKey|TestAccNetworkACLHTTPMessageSignature"

# Against live tenant:
AUTH0_DOMAIN=... AUTH0_CLIENT_ID=... AUTH0_CLIENT_SECRET=... \
  make test-acc-e2e FILTER="TestAccNetworkACLKey|TestAccNetworkACLHTTPMessageSignature"
```

**`TestAccNetworkACLKey`** — create key, verify computed fields (`fingerprint`, `created_at`, `updated_at`); change `value` triggers replacement (fingerprint diff); import leaves `value` absent from state.

**`TestAccNetworkACLKeyValidation`** — invalid `alg` and short `value` (< 32 decoded bytes) are rejected client-side.

**`TestAccNetworkACLHTTPMessageSignature`** — create ACL with one key reference, add a second key, import round-trip.

### 📝 Checklist


internal/auth0/networkaclkey/resource.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	if len(decoded) > 512 {
+		return diag.Errorf("decoded key material must be at most 512 bytes, got %d", len(decoded))
+	}
+
+	return nil
+}
+
+func customizeDiffNetworkACLKey(_ context.Context, d *schema.ResourceDiff, _ interface{}) error {
```
bkiran6398, line 128: `value` parameter with `Sensitive: true` is still persisted in the state. Only `WriteOnly` parameters are excluded from the state file, so a custom diff with additional fingerprint calculation may not be necessary.

internal/auth0/networkaclkey/data_source.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func readNetworkACLKeyDataSource(ctx context.Context, data *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	apiv3 := meta.(*config.Config).GetAPIV3()
+
+	id := data.Get("id").(string)
+
+	key, err := apiv3.Keys.NetworkACLs.Get(ctx, id)
+	if err != nil {
+		return internalError.HandleReadAPIError("auth0_network_acl_key", data, err)
```
duedares-rvj: This is not supposed to be used for Datasource. 
Should be `return diag.FromErr(err)`

internal/auth0/networkaclkey/resource.go
```diff
@@ -0,0 +1,234 @@
...
+
+func readNetworkACLKey(ctx context.Context, data *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	apiv3 := meta.(*config.Config).GetAPIV3()
+
+	key, err := apiv3.Keys.NetworkACLs.Get(ctx, data.Id())
+	if err != nil {
+		return internalError.HandleReadAPIError("auth0_network_acl_key", data, err)
+	}
+
+	// If the API fingerprint differs from what is in state, the key was rotated out-of-band.
```
duedares-rvj, line 179: This is fine. Consider the below approach:

```
return internalError.RemoveFromStateWithWarning("auth0_network_acl_key", data, "the key material fingerprint changed — the key may have been rotated out of band")
```

ref: readConnectionClient implementation.

internal/auth0/networkaclkey/resource.go
```diff
@@ -0,0 +1,234 @@
...
+	var conflictErr *managementv3.ConflictError
+	if errors.As(err, &conflictErr) {
+		return diag.Errorf(
+			"cannot delete network ACL key %q: it is still referenced by one or more ACL rules. "+
+				"Remove the key references from all ACL rules before deleting. API response: %v",
+			data.Id(), conflictErr.Body,
+		)
+	}
+
+	if isHTTPStatus(err, http.StatusConflict) {
```
duedares-rvj: Seems like both are 409 checks. Consider removing redundant checks.

internal/auth0/networkaclkey/resource.go
KartikJha: Updarted

internal/auth0/networkaclkey/data_source.go
KartikJha: Updated

internal/auth0/networkaclkey/resource.go
KartikJha: Updated, but we still need fingerprint customize diff to detect change

internal/auth0/networkaclkey/resource.go
KartikJha: Updated

bkiran6398 (approved):

PR #1699, opened 2026-09-04
feat: add b2b integration configuration to client (EA only)

bkiran6398:
### 🔧 Changes

Adds a `b2b_integration_configuration` block to the `auth0_client` resource (and the derived data source) for Enterprise Connect, an Early Access feature.

- `integration_type` accepts `custom_auth_server`, `third_party`, or `application`.
- `sso_profiles` links a self-service SSO profile id (`ssp_...`), capped at one entry.
- The block is set at create time: adding or removing it forces client recreation, while its contents update in place. An empty block is rejected by the API.
- `organization_discovery_methods` becomes optional and computed so the server default applied to these clients is absorbed without a perpetual diff. 
- **Note: Change in behavior of `organization_discovery_methods`** - Omitting the attribute now retains the current value; should be set to `[]` to clear.

```hcl
resource "auth0_client" "b2b" {
  name = "My B2B App"

  b2b_integration_configuration {
    integration_type = "third_party"
    sso_profiles     = [auth0_self_service_profile.example.id]
  }
}
```

### 📚 References

- Depends on https://github.com/auth0/go-auth0/pull/863

### 🔬 Testing

- Unit tests for flattening the new block.
- Recorded acceptance tests: create per `integration_type`, in-place content update, import, block-removal recreation, and an `sso_profiles` create-then-swap case.
- Manual verification against a live tenant: full CRUD, `organization_discovery_methods` retain-on-omit and clear-on-`[]` for both b2b and non-b2b clients, empty-block rejection, plan-time `sso_profiles` cap, data-source read, and import with an unrelated update.

### 📝 Checklist


internal/auth0/client/resource.go
```diff
@@ -74,6 +81,14 @@ func NewResource() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			StateContext: importClient,
 		},
+		// `b2b_integration_configuration` is immutable after client creation; the Management API rejects adding or clearing it later.
```
duedares-rvj, line 84: Spec mentioned that null can clear this, is that outdated information?

internal/auth0/client/resource.go
```diff
@@ -251,11 +266,40 @@ func NewResource() *schema.Resource {
 					}, false),
 				},
 				Optional: true,
+				Computed: true,
```
duedares-rvj, line 269: Soft breaking change? Since our isNull logic would earlier set this to null when the field is removed. +1 on adding relevant comments on the description. Can accept this change if there's no better way forward.

go.mod
```diff
@@ -4,8 +4,8 @@ go 1.25.8
 
 require (
 	github.com/PuerkitoBio/rehttp v1.4.0
-	github.com/auth0/go-auth0 v1.48.1-0.20260902130806-3d6d163b4162
-	github.com/auth0/go-auth0/v3 v3.4.0
```
duedares-rvj, line 8: 3.4.0 should also work right? Just curious.

internal/auth0/client/resource.go
bkiran6398: Seems like an outdated information in spec.

**Request**
```http
PATCH /api/v2/clients/2WyOlpyjwrrf7WgrZQOEQkMqTQQUu6Iw
Authorization: Bearer <redacted>
Content-Type: application/json

{"b2b_integration_configuration":null}
```

**Response — HTTP 400**
```json
{
  "statusCode": 400,
  "error": "Bad Request",
  "message": "Payload validation error: 'Expected type object but found type null' on property b2b_integration_configuration (Configuration for B2B Integration clients).",
  "errorCode": "invalid_body"
}
```

internal/auth0/client/resource.go
bkiran6398: Unfortunately, there isn't any better way to accommodate backend default value for `organization_discovery_methods`

go.mod
bkiran6398: Yes, should absolutely work fine with v3.4.0. Seems like a miss during conflict resolution. Will fix this.

duedares-rvj (approved):

PR #1702, opened 2026-09-07
feat: unconditional deny init

KartikJha:
### 🔧 Changes

Adds a `match_all` boolean field to the `rule` block of the `auth0_network_acl` resource (and the derived data source) to support unconditional-deny rules — a tenant ACL that blocks or allows all traffic without any signal-based criteria.

- `match_all = true` sends `{ rule: { match_all: true } }` to the API and is mutually exclusive with `match` and `not_match` blocks (plan-time `ConflictsWith` error if combined).
- `match_all = false` is treated as absent — the API only accepts `true`, so the provider never sends `false` in the payload.
- `AtLeastOneOf` on `match` and `not_match` is updated to include `rule.0.match_all`, so a rule with only `match_all = true` passes schema validation.
- An expand-time guard returns a clear error when `match_all` is `false` and no `match`/`not_match` block is provided — covering the edge case where TF SDK v2 may treat `false` as satisfying `AtLeastOneOf`.

```hcl
resource "auth0_network_acl" "deny_all" {
  description = "Block all tenant traffic"
  active      = true
  priority    = 1

  rule {
    action { block = true }
    scope     = "tenant"
    match_all = true
  }
}
```

### 📚 References

- Requires go-auth0 v1 backport: `MatchAll *bool` on `NetworkACLRule` — already merged and pinned at pseudo-version `v1.48.1-0.20260907034825-f34585e28ba4`.

### 🔬 Testing

- **Unit tests** (`expand_flatten_test.go`): 5 cases covering expand with `match_all=true`, expand with `match_all=false`+signal, expand guard with no signals, flatten with `MatchAll=*true`, and flatten with `MatchAll=nil`.
- **Recorded acceptance tests** (`TestAccNetworkACLMatchAll`): create with `match_all=true`, import round-trip (no plan diff), update match_all→signal, update signal→match_all.
- **Recorded conflict test** (`TestAccNetworkACLMatchAllConflict`): verifies plan-time rejection when `match_all=true` and `match` are both set.
- All tests ran against `kartik-esd-57610.sus.auth0.com` and recordings committed.

### 📝 Checklist


internal/auth0/networkacl/expand.go
```diff
@@ -87,6 +88,21 @@ func expandNetworkACL(data *schema.ResourceData) (*management.NetworkACL, error)
...
 
+	if matchAll, ok := rule["match_all"].(bool); ok && matchAll {
+		networkACL.Rule.MatchAll = auth0.Bool(true)
+	}
+
+	hasMatch := len(rule["match"].([]interface{})) > 0
+	hasNotMatch := len(rule["not_match"].([]interface{})) > 0
+	matchAll, _ := rule["match_all"].(bool)
+
+	if !matchAll && !hasMatch && !hasNotMatch {
```
bkiran6398: Since schema has `AtLeastOneOf` and `ConflictsWith` for these params, this check seems redundant.

bkiran6398:
### Code review

Found 1 issue:

1. Setting `match_all = false` explicitly alongside a `match`/`not_match` block fails at plan time with a spurious `"match": conflicts with rule.0.match_all` error, contradicting this field's own description ("Setting this to false is equivalent to omitting it"). `ConflictsWith` is added to `match`/`not_match` targeting `rule.0.match_all`, but terraform-plugin-sdk v2 treats a present zero-value bool as configured: `validateConflictingAttributes` errors whenever `c.Get("rule.0.match_all")` returns `ok == true`, and `ResourceConfig.get` returns `ok == true` for any key present in config, including one set to `false`. So a user who follows the documented "false is equivalent to omitting it" guidance and writes `match_all = false` next to a signal block is blocked at plan. Consider dropping the schema-level `ConflictsWith` for this pair and relying solely on the expand-time guard (which already reads the bool value), or gate the conflict on `match_all == true`.

https://github.com/auth0/terraform-provider-auth0/blob/c6ab03cb42cc9636c0603c79b09d2c65359fc81f/internal/auth0/networkacl/resource.go#L66-L87

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

internal/auth0/networkacl/expand.go
KartikJha: Updated

KartikJha:
> ### Code review
> Found 1 issue:
> 
> 1. Setting `match_all = false` explicitly alongside a `match`/`not_match` block fails at plan time with a spurious `"match": conflicts with rule.0.match_all` error, contradicting this field's own description ("Setting this to false is equivalent to omitting it"). `ConflictsWith` is added to `match`/`not_match` targeting `rule.0.match_all`, but terraform-plugin-sdk v2 treats a present zero-value bool as configured: `validateConflictingAttributes` errors whenever `c.Get("rule.0.match_all")` returns `ok == true`, and `ResourceConfig.get` returns `ok == true` for any key present in config, including one set to `false`. So a user who follows the documented "false is equivalent to omitting it" guidance and writes `match_all = false` next to a signal block is blocked at plan. Consider dropping the schema-level `ConflictsWith` for this pair and relying solely on the expand-time guard (which already reads the bool value), or gate the conflict on `match_all == true`.
> 
> https://github.com/auth0/terraform-provider-auth0/blob/c6ab03cb42cc9636c0603c79b09d2c65359fc81f/internal/auth0/networkacl/resource.go#L66-L87
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

Updated

bkiran6398 (approved):

PR #1716, opened 2026-09-17
feat: adaptive mfa init commit

KartikJha:
### 🔧 Changes

Makes Auth0 Adaptive MFA 403 `insufficient_entitlement` responses non-fatal across three resources — downgraded to a Terraform
warning so `apply` succeeds on tenants without the add-on, rather than failing the plan entirely.

**Shared entitlement helpers (`internal/error/`)**
- `IsInsufficientEntitlement(err error) bool` — unified checker that handles both the v1 SDK (`management.Error` with
  `Status()==403` and `Code()=="insufficient_entitlement"`) and v3 SDK (`*managementv3.ForbiddenError`). The v1 path required
  a `go-auth0` v1 SDK bump to add `Code() string` to `management.Error` (see `go.mod`); existing v3 callers are unaffected.
- `EntitlementWarning(feature, consequence string) diag.Diagnostic` — shared `diag.Warning` factory with consistent messaging and
  godot-compliant copy.
- `EntitlementReadConsequence` / `EntitlementUpdateConsequence` — exported constants for the standard consequence phrases, used
  across all resource call sites to keep messaging consistent.
- `auth0_attack_protection` migrated to use the shared helpers (removes local `entitlementWarning`, `entitlementReadConsequence`,
  `entitlementUpdateConsequence`).

**`auth0_risk_assessments`**
- Read and Update now check `IsInsufficientEntitlement` before returning an error. On 403, a warning is appended and
  `readRiskAssessmentSettings` is called unconditionally so state reflects API reality.

**`auth0_risk_assessments_new_device`**
- Same pattern. The entitlement check fires before the `fmt.Errorf` wrap so `errors.As` can reach the raw SDK error type.
  `readRiskAssessmentNewDeviceSettings` is called after the warning, consistent with `updateGuardian`.

**`auth0_guardian`**
- `updatePolicy` (the only entitlement-gated call) is pulled out of the `multierror.Append` batch — `go-multierror` does not
  implement `Unwrap() []error`, so `errors.As` cannot traverse into wrapped errors. When `updatePolicy` returns a 403, a warning is
  appended and the remaining 8 factor updaters still run. Non-entitlement errors still return immediately. `readGuardian` is called
  unconditionally at the end so state reflects the API's actual policy value.

### 🔬 Testing

Acceptance tests added for all three resources, each with a VCR cassette returning `403 insufficient_entitlement`.

```bash
make test-acc FILTER="TestAccRiskAssessmentInsufficientEntitlement|TestAccRiskAssessmentNewDeviceInsufficientEntitlement|TestAccGuardianPolicyConfidenceScoreInsufficientEntitlement"
```

**`TestAccRiskAssessmentInsufficientEntitlement`** — PATCH→403 emits warning; read is called after update, subsequent GETs→403
also emit warning; apply succeeds with non-empty plan.

**`TestAccRiskAssessmentNewDeviceInsufficientEntitlement`** — same pattern for `/api/v2/risk-assessments/settings/new-device`.

**`TestAccGuardianPolicyConfidenceScoreInsufficientEntitlement`** — PUT `/guardian/policies`→403 emits warning; `ExpectNonEmptyPlan: true`
and `TestCheckResourceAttr("policy", "never")` assert state reflects API's actual value, not the requested `confidence-score`.

Unit tests in `internal/error/api_error_test.go` cover `IsInsufficientEntitlement` (v1 path, v3 path, nil, scope-403 false-positive)
and `EntitlementWarning` (summary/detail shape).

### 📝 Checklist


internal/auth0/attackprotection/resource.go
```diff
@@ -655,29 +655,6 @@ func validateCaptchaProviderSecrets() schema.CustomizeDiffFunc {
 	}
 }
 
-// Consequences of a missing entitlement, phrased per operation: a read only
-// failed to fetch the remote configuration, whereas an update failed to write
-// the configured values.
-const (
```
bkiran6398, line 661: Instead of completely removing these constants and repeating raw text on every function call, can we move these to `api_error.go` along with `EntitlementWarning` and reuse in function calls.

internal/auth0/riskassessment/resource.go
```diff
@@ -65,6 +67,9 @@ func updateRiskAssessmentSettings(ctx context.Context, data *schema.ResourceData
 	}
 
 	if err := api.RiskAssessment.UpdateSettings(ctx, setting); err != nil {
+		if apierr.IsInsufficientEntitlement(err) {
```
bkiran6398, line 72: Similar to `updateGuardian` of `internal/auth0/guardian/resource.go`, append the warning to diags and allow calling `readRiskAssessmentSettings` in return.

harshithRai (approved):

PR #1729, opened 2026-09-21
fix: drop unsupported windows/arm build target

duedares-rvj:
## Summary

- Go 1.26 (introduced via #1709) dropped support for `windows/arm` (32-bit ARM)
- The `.goreleaser.yml` build matrix included `arm` in `goarch` with no ignore rule for `windows/arm`, causing the v1.58.0 release pipeline to fail
- Adds `windows/arm` to the ignore list, matching the existing `darwin/386` pattern


KartikJha (approved):
LGTM

